PR #14, opened 2017-05-26
Remove dependency on wget

webwurst:
This script uses `curl` and `wget`. I just found out when trying this in an environment with missing `wget`. And `curl` might be a bit more common.

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -317,7 +317,7 @@ function dind::ensure-downloaded-kubectl {
   local path="${KUBECTL_DIR}/${link_target}"
   if [[ ! -f "${path}" ]]; then
     mkdir -p "${KUBECTL_DIR}"
-    wget -O "${path}" "https://storage.googleapis.com/kubernetes-release/release/${full_kubectl_version}/bin/${kubectl_os}/amd64/kubectl"
+    curl -o "${path}" "https://storage.googleapis.com/kubernetes-release/release/${full_kubectl_version}/bin/${kubectl_os}/amd64/kubectl"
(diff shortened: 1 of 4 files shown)


dind-cluster.sh
```diff
@@ -317,7 +317,7 @@ function dind::ensure-downloaded-kubectl {
   local path="${KUBECTL_DIR}/${link_target}"
   if [[ ! -f "${path}" ]]; then
     mkdir -p "${KUBECTL_DIR}"
-    wget -O "${path}" "https://storage.googleapis.com/kubernetes-release/release/${full_kubectl_version}/bin/${kubectl_os}/amd64/kubectl"
+    curl -o "${path}" "https://storage.googleapis.com/kubernetes-release/release/${full_kubectl_version}/bin/${kubectl_os}/amd64/kubectl"
```
holser: I would add
-s - silent
-S - show errors 
so the command should be
**curl -sS -o** 
or
**curl -sSo**

but even now it looks good.

holser: Also I found wget in

> fixed/dind-cluster-v1.6.sh
> fixed/dind-cluster-v1.5.sh
> fixed/dind-cluster-v1.4.sh
> image/wrapkubeadm

Could we fix them also? Thank you in advance.

pigmej: @holser @webwurst   fixed are autogenerated ( https://github.com/Mirantis/kubeadm-dind-cluster/blob/master/build/genfixed.sh ) but here it may be even easier to replace them without generating ;)

ivan4th:
LGTM, thanks!

PR #24, opened 2017-07-04
Fix tempfile removal

marcosnils:
I guess the idea was to remove the tempfile here?


ivan4th:
Thanks for the PR! Good catch. Wonder how I managed to type this :)

marcosnils:
> Thanks for the PR! Good catch. Wonder how I managed to type this :)

it happens to us all!. Thx for this awesome project, keep up the good work!. I'll try to contribute more if I find the time.

PR #26, opened 2017-07-06
Upgrade to Kubernetes 1.7, drop k8s 1.4

### LINKED ISSUE #25: support Kubernetes v1.7
I have not looked yet the extent to which a new script and changes would be needed to support v1.7 but putting a placeholder here so that it can be tracked if that's OK.

ivan4th:
Fixes #25.


idvoretskyi:
@ivan4th long-awaited, thanks!

PR #31, opened 2017-07-20
Allow customization of gce-setup.sh and fix dind-cluster.sh bug

pmichali:
It looks like e2e.go has changed the -check_version_skew argument to
-check-version-skew. Changed the dind-cluster-v1.7.sh too.

Allowing user to override VM name, so that gce-setup.sh could be used
from multiple machines at once.  Allowing zone to be overridden by
user,a nd if not, will use gcloud config to set zone. This supports
developers running from different locations.

### DIFF
--- a/gce-setup.sh
+++ b/gce-setup.sh
@@ -20,6 +20,7 @@ else
 fi
 DIND_ROOT="$(cd $(dirname "$(readlinkf "${BASH_SOURCE}")"); pwd)"
 KUBE_DIND_GCE_PROJECT="${KUBE_DIND_GCE_PROJECT:-$(gcloud config list --format 'value(core.project)' 2>/dev/null)}"
+KUBE_DIND_GCE_ZONE="${KUBE_DIND_GCE_ZONE:-$(gcloud config list --format 'value(compute.zone)' 2>/dev/null)}"
 # Based on instructions from k8s build/README.md
 if [ -z "${KUBE_DIND_GCE_PROJECT:-}" ]; then
     echo >&2 "Please set KUBE_DIND_GCE_PROJECT or use 'gcloud config set project NAME'"
@@ -27,14 +28,14 @@ if [ -z "${KUBE_DIND_GCE_PROJECT:-}" ]; then
 fi
 
 set -x
-KUBE_DIND_VM=k8s-dind
+KUBE_DIND_VM="${KUBE_DIND_VM:-k8s-dind}"
 export KUBE_RSYNC_PORT=8730
 export APISERVER_PORT=8899
 docker-machine create \
                --driver=google \
                --google-project=${KUBE_DIND_GCE_PROJECT} \
                --google-machine-image=ubuntu-os-cloud/global/images/ubuntu-1604-xenial-v20170307 \
-               --google-zone=us-west1-a \
+               --google-zone=${KUBE_DIND_GCE_ZONE} \
                --google-machine-type=n1-standard-8 \
                --google-disk-size=50 \
                --google-disk-type=pd-ssd \


gce-setup.sh
```diff
@@ -20,21 +20,22 @@ else
 fi
 DIND_ROOT="$(cd $(dirname "$(readlinkf "${BASH_SOURCE}")"); pwd)"
 KUBE_DIND_GCE_PROJECT="${KUBE_DIND_GCE_PROJECT:-$(gcloud config list --format 'value(core.project)' 2>/dev/null)}"
+KUBE_DIND_GCE_ZONE="${KUBE_DIND_GCE_ZONE:-$(gcloud config list --format 'value(compute.zone)' 2>/dev/null)}"
```
pmichali: Is it possible that test suite doesn't have a zone configured in gcloud? Do we need to check for that condition?

ivan4th:
Hmm, I have problems reproducing the CI issue outside Travis env... Perhaps it has to deal with Docker version. Will try a bit more

pmichali:
@ivan4th Des the Travis-ci run the test.sh script found in the repo?

BTW, I'm using native Mac (but have used Ubuntu 16.04 in a VM too). Both are using 17.06.0-ce for docker. Using Go 1.8.3.

I tried running test.sh in the Ubuntu 16.04 VM. The 1.6 test passed. The src-1.7 had a timeout on a request related to node 2 start. The src-master hung waiting for DNS and dashboard to respond.

I think these errors are unrelated to the changeset, which sets region for GCE and sets the check-version-skew flag, which, if wrong, would have failed with an arg error and not exhibit the issue seen. Advice?

pmichali:
I ran the test.sh script under native Mac. It looks like all the tests passed, except "test-case-src-master". Looks like a bug in the test script. In "test-case-src-1.7" it calls "test-cluster-src", which clones the kubernetes repo. Later, "test-case-src-master" runs and also calls "test-cluster-src" and fails, because the directory already exists.  I'll fix this (surprised it hasn't been seen), and push as part of this commit.

pmichali:
@ivan4th Could use some help on figuring out what is going on here. With this change set, I'm making GCE changes (which are not part of the test.sh run), fixing check-version-skew arg syntax (which would hard fail, if not correct), and recently, changing to allow testing to continue, when kubernetes repo has been already pulled from a previous test (again, would hard fail).

Running locally, I've been able to get every test to pass, except test-master, which fails to get a good health check on either DNS or dashboard containers.

In the latest CI run, I see that test src-1.7 failed. There, it looks like health check failing, on configmap(?), and the test times out.

Test src-master also failed, in this case, starting kubemaster it timed out waiting for control plane. However, I do see that the control plane DID get ready in 45 secs. Seems like maybe a test timing issue? On my local testbed, the control plane came up in 40 seconds.

On a curious note, the test.sh script checks several plugins for 1.5, 1.6, and 1.7 (along with two source checks). Do we need to support 1.5? Just wondering, as it takes really long to run the tests.

Lastly, do you know if the tests pass on the latest source versions w/o any changes to the DinD infrastructure? I'm concerned that there is a problem with the baseline testing with sources.

I've got another change to DinD that I want to push up for review, but am hoping we have a stable base to work with.

pmichali:
@ivan4th @mattymo I see that the check_version_skew option was removed from the code recently. Should I pull that part from this commit and resubmit?

pmichali:
I rebased w/o the fix to check-version-skew, since there was a commit to address that.

ivan4th:
I tried this version of `gce-setup.sh` with these changes:
```
diff --git a/gce-setup.sh b/gce-setup.sh
index c05c31c..05b359a 100644
--- a/gce-setup.sh
+++ b/gce-setup.sh
@@ -34,17 +34,18 @@ export APISERVER_PORT=8899
 docker-machine create \
                --driver=google \
                --google-project=${KUBE_DIND_GCE_PROJECT} \
-               --google-machine-image=ubuntu-os-cloud/global/images/ubuntu-1604-xenial-v20170307 \
+               --google-machine-image=ubuntu-os-cloud/global/images/ubuntu-1604-xenial-v20170803 \
                --google-zone=${KUBE_DIND_GCE_ZONE} \
                --google-machine-type=n1-standard-8 \
                --google-disk-size=50 \
                --google-disk-type=pd-ssd \
                --engine-storage-driver=overlay2 \
+               --engine-install-url=https://releases.rancher.com/install-docker/17.06.sh \
                ${KUBE_DIND_VM}
 eval $(docker-machine env ${KUBE_DIND_VM})
 docker-machine ssh ${KUBE_DIND_VM} \
                -L ${KUBE_RSYNC_PORT}:localhost:${KUBE_RSYNC_PORT} \
                -L ${APISERVER_PORT}:localhost:${APISERVER_PORT} \
                -N&
-time "${DIND_ROOT}"/dind-cluster.sh up
+# time "${DIND_ROOT}"/dind-cluster.sh up
 set +x
```

and `TEST_CASE=src-1.7 bash -x ./test.sh` succeeded after that (note that `gce-setup.sh` is meant to be an example). As of `TEST_CASE=src-master` and Travis failures on `src-*` tests -- looking into that.

(the changes above are about using newer ubuntu image and working around docker-machine problem on mac docker + not starting the cluster before running test.sh)

pmichali:
@ivan4th Thanks about the point of gce-setup.sh being an example. We're thinking that it could be useful, when using DInD under a CI suite to test IPv6 with Kubernetes. It would be nice, if we can have some options to make it more flexible for people and to fix into CI use.

Since the source tests have been disabled, I'm not so worried (as this change doesn't really affect DinD operation, is just a wrapper), so if we can proceed with this that would be great.

In general though, I find the source test mode useful, especially with the PR option and working with master.  It might be nice to have an option on the test.sh script that would bring everything up and stop, allowing people to use the system as-is, and then manually shutting down the cluster.

ivan4th:
Well, re-checked -- this PR also worked for me as-is, so merging. Concerning enabling src tests on Travis: I need to debug this elusive Travis issue, I've requested debug mode for Travis builds of k-d-c and now that I have it granted I can try ssh into the VMs and see what's wrong. As of adding an option for `test.sh` to just set up the env -- I'll do this.

pmichali:
@ivan4th Awesome! Thanks for the review and looking forward to the upcoming changes.

PR #35, opened 2017-08-03
Don't fail if kubernetes repo pulled already.

pmichali:
This fixes the issue where two tests (src-1.7 and src-master) both try
to clone the kubernetes repo, and the latter fails, because the repo
already exists. Seen in local testing.

A secondary purpose of this commit is to check the baseline CI testing
for DinD (s it doesn't change any DinD code). There is a concern that
there may be problems with the two src test cases in test.sh.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -66,7 +66,9 @@ function test-cluster {
 function test-cluster-src {
   (
     local version="${1:-}"
-    git clone https://github.com/kubernetes/kubernetes.git
+    if [[ ! -d "kubernetes" ]]; then
+       git clone https://github.com/kubernetes/kubernetes.git
+    fi
     cd kubernetes
     if [[ ${version} ]]; then
       git checkout "${version}"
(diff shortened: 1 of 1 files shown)


pmichali:
@ivan4th Looks like there is an issue with the baseline for src-1.7 and src-master tests, as this change doesn't affect the operation of those two test cases. Can you please advise on how we can get the tests passing? I've got several commits that I'd like to upstream.

pmichali:
@idvoretskyi @djschny @ivan4th @webwurst  Looking at this locally, I can reproduce a failure, when running with TEST_CASE=src-master. What I see is that dind-cluster.sh is joining the minion nodes (they are created), and that fails. Later, it tries to start up the dashboard, and that fails, because the only node is kube-master which has NoSchedule.  I tried to manually join the node and see an error that the overlay2 driver is not supported. Not sure why - the node seems to be alpine OS.

So, there is some issue with the baseline K8s code running under DinD.

pmichali:
@ivan4th Could use some help here. I think this change is needed for source based testing, however I see that the source tests have been recently pulled from test.sh.  Should I rebase and commit these changes?

Also, are there plans to get the source based test running?

pmichali:
I did a rebase and am only including this change to the source based tests.

ivan4th:
Ok, thanks for the PR!

PR #42, opened 2017-10-03
Provide IPv6 support

pmichali:
Adding a WIP change (on top of the fix-1.8+ branch) that will allow DinD to be used on an IPv6 only Kubernetes cluster.

This uses IPv6 addresses for all pods, and makes use of DNS64 and NAT64 containers, to convert external IPs to IPv4 embedded IPv6 addresses.

An IP_MODE environment variable is added to select between IPv4 (default) and IPv6. Additional environment variables can be set to customize DinD subnet, service subnet CIDR, and prefix to use for DNS64.

To use, must use the customer Kubernetes repo in the README.md, as it has IPv6 patches that are in-flight (or fixing bugs).

Notes:
- Uses /80 for pod subnet size.
- Requires building local copy, due to changes to DinD code.
- All pod subnets are "inside" of the DinD /64 subnet (same as what IPv4 does).
- Includes fix to restore controller manifest, when failure to shutdown controller/proxy pods during snapshot.
- Relies on both a custom Kubernetes repo (until commits are upstreamed), and kube-dns image (referenced in manifest of custom Kubernetes repo).
- Could be customized to use NAT66 container as an option to allow access to external IPv6 IPs.
- Future could add dual-stack support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,105 @@ The first `dind/dind-cluster.sh up` invocation can be slow because it
 needs to build the base image and Kubernetes binaries. Subsequent
 invocations are much faster.
 
+## IPv6 Mode (experimental)
+To run Kubernetes in IPv6 only mode, the following steps can be performed
+in a Kubernetes source setup. First, build a new kubeadm-dind-cluster
+image, from the DinD area:
+
+```shell
+$ cd ~/dind
+
+$ build/build-local.sh
+$ export DIND_IMAGE=mirantis/kubeadm-dind-cluster:local
+```
+
+Next, clone a Kubernetes repo from master branch:
+
+```shell
+$ git clone https://github.com/kubernetes/kubernetes.git
+$ cd kubernetes
+```
+
+As of Oct. 23rd, 2017, the following IPv6 PRs are in-flight, and should
+be cherry picked and added to the repo:
+
+```
+PR #53484 "Adds Support for Configurable Kubeadm Probes."
+PR #45551 "Adds Support for Node Resource IPv6 Addressing"
+PR #53148 "ip6tables should be set in the noop plugin"
+PR #52180 "Updating kubenet for CNI with IPv6"
+PR #45792 "Updating NewCIDRSet return a value"
+PR #50929 "Add kubeadm config for setting kube-proxy bind address"
+PR #53555 "Add IPv6 and negative UT test cases for proxier's deleteEndpointConnections"
+PR #52748 "Add brackets around IPv6 addrs in e2e test IP:port endpoints"
+PR #54639 "Updates kube-dns in kubeadm to 1.14.7"
+PR #54437 "Adds support for v4/v6 loopback dns bind address."
+```
+
+These can be obtained with doing cherry picks from the PR commits:
+```
+git fetch origin pull/53484/head:pr53484
+git log --abbrev-commit pr53484 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+
+git fetch origin pull/45551/head:pr45551
+git log --abbrev-commit pr45551 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+
+git fetch origin pull/53184/head:pr53184
+git log --abbrev-commit pr53184 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+
+git fetch origin pull/52180/head:pr52180
+git log --abbrev-commit pr52180 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+
+git fetch origin pull/45792/head:pr45792
+git log --abbrev-commit pr45792 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+```
+
+Had conflict on above and needed to resolve.
+
+```
+git fetch origin pull/50929/head:pr50929
+git log --abbrev-commit pr50929 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+
+git fetch origin pull/53555/head:pr53555
+git log --abbrev-commit pr53555 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+
+git fetch origin pull/52748/head:pr52748
+git log --abbrev-commit pr52748 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+
+git fetch origin pull/54639/head:pr54639
+git log --abbrev-commit pr54639 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+
+git fetch origin pull/54437/head:pr54437
+git log --abbrev-commit pr54437 --oneline --abbrev-commit -n 1 | cut -f 1 -d" "
+```
+
+If you have a host that allows four times more conntrack connection (max) than
+hashsize, You'll need this temporary patch (until issue 50787 is fixed and DinD is
+modified to customize the kubelet config file):
+
+https://github.com/pmichali/kubernetes/commit/d76164d919f292915ebaf69d8a3c357f40457bd7
+
+Lastly, set up environment variables for building and IPv6 mode and
+then bring up the cluster:
+
+```shell
+$ export BUILD_KUBEADM=y
+$ export BUILD_HYPERKUBE=y
+$ export IP_MODE=ipv6
+
+$ ../dind-cluster.sh up
+```
+
+Note: there are additional customizations that you can make for IPv6,
+to set the prefix used for DNS64, subnet prefix to use for DinD, and
+the service subnet CIDR:
+
+```shell
+export DNS64_PREFIX=fd00:77:64:ff9b::
+export DIND_SUBNET=fd00:77::
+export SERVICE_CIDR=fd00:77:30::/110
+```
+
 ## Configuration
 You may edit `config.sh` to override default settings. See comments in
 [the file](config.sh) for more info. In particular, you can specify
--- a/config.sh
+++ b/config.sh
@@ -1,5 +1,10 @@
-# DIND subnet (/16 is always used)
-DIND_SUBNET=10.192.0.0
+if [[ ${IP_MODE} = "ipv4" ]]; then
+    # DinD subnet (expected to be /16)
+    DIND_SUBNET="10.192.0.0"
+else
+    # DinD subnet (expected to be /64)
+    DIND_SUBNET="fd00:10::"
+fi
 
 # Apiserver port
 APISERVER_PORT=${APISERVER_PORT:-8080}
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -42,6 +42,7 @@ fi
 
 #%CONFIG%
 
+IP_MODE="${IP_MODE:-ipv4}"  # ipv4, ipv6, (future) dualstack
 if [[ ! ${EMBEDDED_CONFIG:-} ]]; then
   source "${DIND_ROOT}/config.sh"
 fi
@@ -51,9 +52,42 @@ if [[ CNI_PLUGIN = "calico" || CNI_PLUGIN = "calico-kdd" ]]; then
   DEFAULT_POD_NETWORK_CIDR="192.168.0.0/16"
 fi
 CNI_PLUGIN="${CNI_PLUGIN:-bridge}"
-DIND_SUBNET="${DIND_SUBNET:-10.192.0.0}"
 POD_NETWORK_CIDR="${POD_NETWORK_CIDR:-${DEFAULT_POD_NETWORK_CIDR}}"
-dind_ip_base="$(echo "${DIND_SUBNET}" | sed 's/\.0$//')"
+ETCD_HOST="${ETCD_HOST:-127.0.0.1}"
+GCE_HOSTED="${GCE_HOSTED:-false}"
+if [[ ${IP_MODE} = "ipv6" ]]; then
+    dind_ip_base="${DIND_SUBNET}"
+    ETCD_HOST="::1"
+    KUBE_RSYNC_ADDR="${KUBE_RSYNC_ADDR:-::1}"
+    SERVICE_CIDR="${SERVICE_CIDR:-fd00:10:30::/110}"
+    DIND_SUBNET_SIZE="${DIND_SUBNET_SIZE:-64}"
+    REMOTE_DNS64_V4SERVER="${REMOTE_DNS64_V4SERVER:-8.8.8.8}"
+    LOCAL_NAT64_SERVER="${DIND_SUBNET}200"
+    DNS64_PREFIX="${DNS64_PREFIX:-fd00:10:64:ff9b::}"
+    DNS64_PREFIX_SIZE="${DNS64_PREFIX_SIZE:-96}"
+    DNS64_PREFIX_CIDR="${DNS64_PREFIX}/${DNS64_PREFIX_SIZE}"
+    dns_server="${dind_ip_base}100"
+    # Create a /64 prefix so that created /80 pod net is inside the /64 DinD network
+    POD_NET_PREFIX="$(echo $DIND_SUBNET | sed 's/::$//')"
+    num_colons="$(grep -o ":" <<< "${POD_NET_PREFIX}" | wc -l)"
+    while [ $num_colons -lt 3 ]; do
+	POD_NET_PREFIX+=":0"
+	let num_colons+=1
+    done
+else
+    dind_ip_base="$(echo "${DIND_SUBNET}" | sed 's/0$//')"
+    KUBE_RSYNC_ADDR="${KUBE_RSYNC_ADDR:-127.0.0.1}"
+    SERVICE_CIDR="${SERVICE_CIDR:-10.96.0.0/12}"
+    DIND_SUBNET_SIZE="${DIND_SUBNET_SIZE:-16}"
+    dns_server="${REMOTE_DNS64_V4SERVER:-8.8.8.8}"
+fi
+dns_prefix="$(echo ${SERVICE_CIDR} | sed 's,/.*,,')"
+if [[ ${IP_MODE} != "ipv6" ]]; then
+    dns_prefix="$(echo ${dns_prefix} | sed 's/0$//')"
+fi
+kube_master_ip="${dind_ip_base}2"
+DNS_SVC_IP="${dns_prefix}10"
+
 DIND_IMAGE="${DIND_IMAGE:-}"
 BUILD_KUBEADM="${BUILD_KUBEADM:-}"
 BUILD_HYPERKUBE="${BUILD_HYPERKUBE:-}"
@@ -117,7 +151,7 @@ function dind::set-build-volume-args {
   if [ -n "${KUBEADM_DIND_LOCAL:-}" ]; then
     build_volume_args=(-v "$PWD:/go/src/k8s.io/kubernetes")
   else
-    build_container_name="$(KUBE_ROOT=$PWD &&
+    build_container_name="$(KUBE_ROOT=${PWD} ETCD_HOST=${ETCD_HOST} &&
                             . ${build_tools_dir}/common.sh &&
                             kube::build::verify_prereqs >&2 &&
                             echo "${KUBE_DATA_CONTAINER_NAME:-${KUBE_BUILD_DATA_CONTAINER_NAME}}")"
@@ -372,7 +406,12 @@ function dind::ensure-binaries {
 
 function dind::ensure-network {
   if ! docker network inspect kubeadm-dind-net >&/dev/null; then
-    docker network create --subnet="${DIND_SUBNET}/16" kubeadm-dind-net >/dev/null
+    local v6settings=""
+    if [[ ${IP_MODE} = "ipv6" ]]; then
+      # Need second network for NAT64
+      v6settings="--subnet=172.18.0.0/16 --ipv6"
+    fi
+    docker network create ${v6settings} --subnet="${DIND_SUBNET}/${DIND_SUBNET_SIZE}" --gateway="${dind_ip_base}1" kubeadm-dind-net >/dev/null
   fi
 }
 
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -123,6 +123,48 @@ The first `dind/dind-cluster.sh up` invocation can be slow because it
...
+$ export DIND_IMAGE=mirantis/kubeadm-dind-cluster:local
+```
+
+Next, clone this custom Kubernetes repo, which has a set of patches on
+top of a recent master release:
+
+```shell
+$ git clone https://github.com/pmichali/kubernetes.git
+$ cd kubernetes
+$ git checkout -b kubernetes_ipv6
```
danehans: Instead of using your fork, I suggest using upstream and listing the necessary PRs. As the PRs merge, doc update PRs can be submitted.

README.md
```diff
@@ -123,6 +123,48 @@ The first `dind/dind-cluster.sh up` invocation can be slow because it
...
+## IPv6 Mode (experimental)
+To run Kubernetes in IPv6 only mode, the following steps can be performed
+in a Kubernetes source setup. First, build a new kubeadm-dind-cluster
+image, from the DinD area:
+
+```shell
+$ cd ~/dind
+
+$ build/build-local.sh
+$ export DIND_IMAGE=mirantis/kubeadm-dind-cluster:local
```
danehans: Is this build step needed if this PR gets merged? If so, I suggest removing it.

config.sh
```diff
@@ -1,6 +1,3 @@
-# DIND subnet (/16 is always used)
-DIND_SUBNET=10.192.0.0
-
```
danehans: Why does din subnet need to be removed from the config?

dind-cluster.sh
```diff
@@ -368,7 +402,12 @@ function dind::ensure-binaries {
 
 function dind::ensure-network {
   if ! docker network inspect kubeadm-dind-net >&/dev/null; then
-    docker network create --subnet="${DIND_SUBNET}/16" kubeadm-dind-net >/dev/null
+    local v6settings=""
+    if [[ ${IP_MODE} = "ipv6" ]]; then
+      # Need second network
```
danehans: Can you add to the comment why a 2nd network is needed, i.e. for nat64.

README.md
pmichali: I wanted to have an easy way for people to test, rather than forcing them to patch in a bunch of PRs. I do list the PRs in my blog (http://blog.michali.net/2017/08/09/ipv6-support-for-docker-in-docker/).

Would like to know how we want to manage this. Also, I have one change for a temp hack for an issue where there is no PR yet.

dind-cluster.sh
```diff
@@ -557,12 +666,27 @@ function dind::init {
...
+    local pod_net_cidr=""
+    # TODO: May want to specify each of the plugins that require --pod-network-cidr
+    if [[ ${CNI_PLUGIN} != "bridge" ]]; then
+      pod_net_cidr="podSubnet: \"${POD_NETWORK_CIDR}\""$'\n'"  "
+    fi
+    local bind_address="0.0.0.0"
+    if [[ $IP_MODE = "ipv6" ]]; then
+	bind_address="::"
+    fi
     docker exec -i kube-master /bin/sh -c "cat >/etc/kubeadm.conf" <<EOF
```
pmichali: I'm wondering if we should built a temp file with kubeadm.conf, and then docker cp it to the master, rather than trying to build on the fly.

config.sh
pmichali: Because DinD will force this as the subnet, and we cannot override, unless we manually set the environment variable EMBEDDED_CONFIG. Instead, I remove it from here, as set it in dind-cluster.sh, allowing the user to override, if desired, and to set a default that is IP mode specific.

dind-cluster.sh
pmichali: Sure will do.

dind-cluster.sh
```diff
@@ -46,10 +46,44 @@ if [[ ! ${EMBEDDED_CONFIG:-} ]]; then
...
+else
+    DIND_SUBNET="${DIND_SUBNET:-10.192.0.0}"
+    dind_ip_base="$(echo "${DIND_SUBNET}" | sed 's/0$//')"
+    KUBE_RSYNC_ADDR="${KUBE_RSYNC_ADDR:-127.0.0.1}"
+    SERVICE_CIDR="${SERVICE_CIDR:-10.96.0.0/12}"
+    DIND_SUBNET_SIZE="${DIND_SUBNET_SIZE:-16}"
+    dns_server="${REMOTE_DNS64_V4SERVER:-8.8.8.8}"
+fi
+dns_prefix=$(echo $SERVICE_CIDR | sed 's,/.*,,')
+if [[ $IP_MODE = "ipv6" ]]; then
```
pmichali: Typo here. Should be !=, so that it pulls the zero off prefix (10.96.0.0/12 -> 10.96.0.0 -> **10.96.0.**)

dind-cluster.sh
```diff
@@ -459,7 +561,7 @@ function dind::kubeadm {
   status=0
   # See image/bare/wrapkubeadm.
   # Capturing output is necessary to grab flags for 'kubeadm join'
-  if ! docker exec "${container_id}" wrapkubeadm "$@" 2>&1 | tee /dev/fd/2; then
+  if ! docker exec "${container_id}" wrapkubeadm ${dns_ip} "$@" 2>&1 | tee /dev/fd/2; then
```
pmichali: Unfortunately, this breaks CI, as running the new dind-cluster.sh, but using an older kubeadm-cluster-dind image, the wrapkubeadm on that image cannot deal with another argument. I'm looking into other options, like setting up the needed KUBELET_DNS_ARGS entry for systemd from the dindnet script, instead of passing into wrapkubeadm.

image/wrapkubeadm
```diff
@@ -327,6 +337,9 @@ function dind::do-kubeadm {
   fi
 }
 
+dns_ip=$1
```
pmichali: As mentioned above, we can't rely on passing in a new argument, as old images will not work with new dind-cluster.sh. So, we can't really do the changes on L306-314 here, as we have no easy way to configure the DNS IP on the fly.

README.md
ivan4th: Indeed, after we merge the PR these commands are redundant. It will be possible to use any up-to-date DIND image for IPv6.

image/snapshot
```diff
@@ -42,8 +42,7 @@ function snapshot::prepare {
     while kubectl get pod kube-controller-manager-kube-master -n kube-system -o name 2>/dev/null | grep -q pods*/; do
       if ((--n == 0)); then
         echo "WARNING: kubelet glitch: kube-controller-manager won't stop; pods may 'blink' for some time after restore" >&2
-        systemctl stop kubelet docker
```
ivan4th: It's important to stop kubelet and docker before making the snapshot, otherwise it can turn out to be inconsistent.

dind-cluster.sh
```diff
@@ -856,8 +985,12 @@ function dind::do-run-e2e {
   local parallel="${1:-}"
   local focus="${2:-}"
   local skip="${3:-}"
+  local host=$kube_master_ip
+  if [[ $IP_MODE = "ipv6" ]]; then
```
ivan4th: `[[ ${IP_MODE} = "ipv6" ]]`

dind-cluster.sh
```diff
@@ -856,8 +985,12 @@ function dind::do-run-e2e {
   local parallel="${1:-}"
   local focus="${2:-}"
   local skip="${3:-}"
+  local host=$kube_master_ip
```
ivan4th: host="${kube_master_ip}"

dind-cluster.sh
```diff
@@ -607,7 +734,7 @@ function dind::join {
 }
 
 function dind::escape-e2e-name {
-    sed 's/[]\$*.^|()[]/\\&/g; s/\s\+/\\s+/g' <<< "$1" | tr -d '\n'
```
ivan4th: Why this change?

dind-cluster.sh
```diff
@@ -557,12 +669,27 @@ function dind::init {
     # insecure-bind-address and insecure-bind-port are also overridden by patching
     # apiserver static pod, but someday when k-d-c will only support
     # kubeadm 1.8+ we'll be able to remove that patching
+    local pod_net_cidr=""
+    # TODO: May want to specify each of the plugins that require --pod-network-cidr
+    if [[ ${CNI_PLUGIN} != "bridge" ]]; then
+      pod_net_cidr="podSubnet: \"${POD_NETWORK_CIDR}\""$'\n'"  "
+    fi
+    local bind_address="0.0.0.0"
+    if [[ $IP_MODE = "ipv6" ]]; then
```
ivan4th: `[[ ${IP_MODE} = "ipv6" ]]`

README.md
```diff
@@ -123,6 +123,63 @@ The first `dind/dind-cluster.sh up` invocation can be slow because it
...
+be cherry picked and added to the repo:
+
+```
+PR #47621 "Updates RangeSize error message and tests for IPv6"
+PR #48551 "Add IPv6 support to iptables proxier"
+PR #50929 "Add kubeadm config for setting kube-proxy bind address"
+PR #50478 "Fix kube-proxy to use proper iptables commands for IPv6"
+PR #52028 "Add required family flag for conntrack IPv6 operation"
+PR #52033 "Removed the IPv6 prefix size limit for cluster-cidr"
+```
```
ivan4th: It would be nice to add some handy commands that may help with cherry-picking the PR, such as `git fetch origin pull/PR_ID/head:LOCAL_BRANCH`

config.sh
ivan4th: `EMBEDDED_CONFIG` actually disallows loading of the config:
```
if [[ ! ${EMBEDDED_CONFIG:-} ]]; then
  source "${DIND_ROOT}/config.sh"
fi
```

The idea of `config.sh` is to set env vars for `dind-cluster.sh` (but not `dind-cluster-vXXX.sh`, i.e. ones that use `EMBEDDED_CONFIG`)

dind-cluster.sh
```diff
@@ -390,6 +432,55 @@ function dind::ensure-volume {
...
+		   resystit/bind9:latest >/dev/null
+	    ipv4_addr=$(docker exec bind9 ip addr list eth0 | grep "inet" | awk '$1 == "inet" {print $2}')
+	    docker exec bind9 ip addr del $ipv4_addr dev eth0
+	    docker exec bind9 ip -6 route add $DNS64_PREFIX_CIDR via $LOCAL_NAT64_SERVER
+        fi
+    fi
+}
+
+function dind::ensure-nat {
+    if [[  $IP_MODE = "ipv6" ]]; then
```
ivan4th: please use `[[ ${IP_MODE} = "ipv6" ]]` for the sake of consistency

dind-cluster.sh
```diff
@@ -390,6 +432,55 @@ function dind::ensure-volume {
...
+        if ! docker ps | grep tayga >&/dev/null; then
+            docker run -d --name tayga --hostname tayga --net kubeadm-dind-net --label mirantis.kubeadm_dind_cluster \
+		   --sysctl net.ipv6.conf.all.disable_ipv6=0 --sysctl net.ipv6.conf.all.forwarding=1 \
+		   --privileged=true --ip 172.18.0.200 --ip6 $LOCAL_NAT64_SERVER --dns $REMOTE_DNS64_V4SERVER --dns $dns_server \
+		   -e TAYGA_CONF_PREFIX=$DNS64_PREFIX_CIDR -e TAYGA_CONF_IPV4_ADDR=172.18.0.200 \
+		   danehans/tayga:latest >/dev/null
+	    # TODO Way to add route w/o sudo? Need to check/create, as "clean" may remove route
+	    local route=$(ip route | egrep "^172.18.0.128/25")
```
ivan4th: `local route="$(ip route | egrep "^172.18.0.128/25")"`

dind-cluster.sh
```diff
@@ -390,6 +432,55 @@ function dind::ensure-volume {
...
+	    if [[ -z "$route" ]]; then
```
ivan4th: `[[ ! ${route} ]]`

dind-cluster.sh
```diff
@@ -479,7 +585,11 @@ function dind::kubeadm {
 
 function dind::configure-kubectl {
   dind::step "Setting cluster config"
-  "${kubectl}" config set-cluster dind --server="http://localhost:${APISERVER_PORT}" --insecure-skip-tls-verify=true
+  local host="localhost"
+  if [[ $IP_MODE = "ipv6" ]]; then
```
ivan4th: `[[ ${IP_MODE} = "ipv6" ]]`

ivan4th:
As of now, the builds fail with this:
```
cp: cannot stat '/config-files/20-extra-dns-args.conf': No such file or directory
```

README.md
pmichali: Will add some. Wasn't sure if you wanted the README with that much detail.

config.sh
pmichali: I moved DIND_SUBNET up to dind-cluster.sh, in order to avoid setting EMBEDDED_CONFIG for IPv6 mode (otherwise, it gets set by config.sh to IPv4 address. Will that work for the older scripts too? It only sets, if the value is not set already.

dind-cluster.sh
pmichali: OK

dind-cluster.sh
pmichali: Will update.

dind-cluster.sh
pmichali: Will do for all.

dind-cluster.sh
pmichali: ditto

dind-cluster.sh
pmichali: ditto

dind-cluster.sh
pmichali: We wanted to run a test like this, and with the escaping, it was not forming the right focus/skip in the underlying test run AFAICT:

```
../dind-cluster.sh e2e "Networking|Services" "IPv4|Networking-Performance|Federation|preserve source pod"
```

dind-cluster.sh
pmichali: will do

dind-cluster.sh
pmichali: ditto

image/snapshot
pmichali: Right. Instead of returning, we are doing a break, which falls through to L61 and will do the stop. Should be equivalent operation, right?

image/wrapkubeadm
```diff
@@ -49,7 +49,7 @@ apiserver_streaming_proxy_redirects='.spec.containers[0].command|=.+["--feature-
 # Update kube-proxy CIDR, enable --masquerade-all and disable conntrack (see dind::frob-proxy below)
 function dind::proxy-cidr-and-no-conntrack {
   cluster_cidr="$(ip addr show docker0 | grep -w inet | awk '{ print $2; }')"
-  echo ".items[0].spec.template.spec.containers[0].command |= .+ [\"--cluster-cidr=${cluster_cidr}\", \"--masquerade-all\", \"--conntrack-max=0\", \"--conntrack-max-per-core=0\"]"
+  echo ".items[0].spec.template.spec.containers[0].command |= .+ [\"--cluster-cidr=${cluster_cidr}\", \"--masquerade-all\", \"--conntrack-max=-1\", \"--conntrack-max-per-core=-1\"]"
```
pmichali: Issue 50787 alludes to being able to set these to -1 (rather than hack kube-proxy code). Just tried it, but it is not working. Investigating.

pmichali: Looks like command line args are deprecated. In addition, they are ignored, if there is a config file (there is). Using a Kuberentes code hack to avoid this issue by diabling defaulting logic from changing conntrack values, if they are not specified in config file (they are not), which on some systems causes a problem as hashsize cannot be changed.

In future, will need to configure config file for kube-proxy with the needed values to prevent override (or get issue 25543 fixed).

pmichali:
@ivan4th I updated the change set, based on your comments. Can you recheck?

Can you elaborate on the "build failure"? That file is not used in the building of the image (which is needed to get all the changes). It is created by dindnet, and then applied (copied into systemd area and then service restarted) during wrapkubeadm operation.

Thanks.

README.md
pmichali: Do we want to leave this in for now, and then remove, once the PR is committed, or pull it now?

config.sh
ivan4th: Maybe a better way is to use separate `DIND_SUBNET_6` and `DIND_SUBNET_4` vars? The idea of `config.sh` is to provide the list of available knobs that customize k-d-c behaviour, and removing a var from there makes its usage unclear.

dind-cluster.sh
ivan4th: Ok.

image/snapshot
ivan4th: Yes, sorry.

test.sh
```diff
@@ -257,6 +257,10 @@ function test-case-src-master-weave {
   )
 }
 
+function test-case-src-working-area {
```
ivan4th: What is this used for?

config.sh
pmichali: It seems like config.sh does more of forcing values to defaults, when not customizing, than showing all the available knobs (several are missing). This is important, as it forces one to customize settings to do IPv6 mode.

With the change set, one can do IPv6 mode, by setting the IP_MODE settings, and will get an IPv6 DIND_SUBNET default value, which can still be overridden, if needed.

If we use multiple variable names for the DinD subnet(we'd need 4, net and size for each IP mode), then we have to add conditional logic, based on the IP mode, to the network create, as well. I think splitting the variable up makes it more confusing.

Should we add IP_MODE to config.sh, defaulting to ipv4, and have logic there that sets DIND_SUBNET default, based on IP mode? IOW, move some of the dind_cluster.sh logic into config.sh?

test.sh
pmichali: Intent is to give the developer a way to run the CI test locally, against their customized local DinD and kubernetes repos. For example, for IPv6, there are numerous PRs in-flight, so we have a Kubernetes repo based on master, with a dozen PRs added.  Intent is not to add this as a test-case for a CI run.

pmichali:
I’ll be pushing up a new commit that has several changes. One is to try to address DIND_SUBNET so that this can be used with IPv6 w/o customizing with EMBEDDED_CONFIG. Let me know what you think.

I have test-case-src-working-area as a way for developers to test the CI operation locally, with a custom Kubernetes repo. For example, I have a repo based on master, but with almost a dozen IPv6 in-flight PRs patched in. Intent is not to use this test case for CI, but for development use. Let me know if you want it removed or commented out.

My PR is based on the fix-1.8+ branch. Do I need to do anything, like rebase to a different branch? I wanted the fix-1.8+ changes. Also, my latest code (that I’m looking to update in the PR), has the workaround change you did for Travis. Is that committed anywhere? fix-1.8+?

I’ve been testing my changes and IPv4/IPv6 seem to work just fine locally. For GCE, I made a gcd-setup-local.sh script that does a build, so that a local DinD image can be used. It works for IPv4, but is not working for IPv6. I had to do several changes, so that files are over on the VM. Seems like it is trying to do kubectl command to start up the dashboard pod, and that request is not getting to the kube-master. It was trying to access localhost:8899, but localhost is 127.0.0.1.

With the latest changes, I tried doing SSH forward of ::1 using port 8899, but it is still failing. Not sure what is wrong there.  Would appreciate any thoughts you have on how to get this going.

Regards,

Paul Michali (pcm)

dind-cluster.sh
```diff
@@ -483,7 +597,15 @@ function dind::kubeadm {
 
 function dind::configure-kubectl {
   dind::step "Setting cluster config"
-  "${kubectl}" config set-cluster dind --server="http://localhost:${APISERVER_PORT}" --insecure-skip-tls-verify=true
+  local host="[${dind_ip_base}2]"
+  if [[ "${GCE_HOSTED}" = true ]]; then
+    if [[ "${IP_MODE}" = "ipv4" ]]; then
+      host="localhost"
+    else
+      host="[::1]"
```
pmichali: For IPv6 GCE setup, kubectl commands are not making it to kube-master.

pmichali:
Have both IPv4 and IPv6 working for bare-metal and GCE configurations.  Removed WIP from title. Please review.

dind-cluster.sh
pmichali: works now.

pmichali:
@ivan4th @danehans Please review so we can upstream this change. Thanks!

pmichali:
Thanks for reviewing and approving @ivan4th !

danehans:
Paul,

Do I call gce-setup-local.sh or gce-setup.sh to deploy to gce with ipv6?

-Daneyon

pmichali:
I only tested with gce-setup-local.sh, because the image for IPv6 was not released. To do that, I needed to set DIND_CLUSTER and BUILD_* flags.  The key here is that the GCE_HOSTED flag needs to be set. If the IPv6 version is now released, we can revise gcd-setup.sh to set the GCE_HOSTED flag and then it would work for IPv6 too. Note: for testing, you could just set the environment variable, and then run gcd-setup.sh and it should work (provided the released image includes the IPv6 stuff).

So, for IPv6 we need:

export BUILD_KUBEADM=y
export BUILD_HYPERKUBE=y
export DIND_IMAGE=mirantis/kubeadm-dind-cluster:local
export IP_MODE=ipv6
. ../gce-setup-local.sh

Or (until updated)…

export GCE_HOSTED=true
export IP_MODE=ipv6
. ../gce-setup.sh

Assuming your Kubernetes area is directly below the DinD directory.

Regards,

Paul Michali (pcm)

Issue #43, opened 2017-10-24
Is there any way to expose NodePort to the host?


pbochynski:
I see that master node has exposed port 8080. I need to access some services exposed with NodePort from my machine. Is there an easy way to enable it?

ReSearchITEng:
Hello,
I am also interested in a way to solve this. 
More exactly, I am interested in deploying nginx ingress controller, but, NodePort does not listen on 127.0.0.1...

```
curl https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get | bash
helm delete --purge nginx-ingress
helm install --name nginx-ingress --namespace kube-system stable/nginx-ingress --set rbac.create=true --set controller.stats.enabled=true --set controller.service.type=NodePort --set controller.service.nodePorts.http=32080 --set controller.service.nodePorts.https=32443 --set controller.stats.enabled=true
#--set controller.service.externalIPs=127.0.0.1
```

If the nodePort would be listening on 127.0.0.1, we could easy expose the ports listening inside the cluster on localhost (see https://github.com/Mirantis/kubeadm-dind-cluster/pull/55 ).

xenophenes:
This issue was never addressed. I'm also interested in how to go about this.

savamilovanovic:
I was able to expose node port using environment variable EXTRA_PORTS.

lukaszo:
NodePort works just fine: 
```bash
$ kubectl get pod
NAME                       READY     STATUS    RESTARTS   AGE
my-nginx-9d5677d94-rv7rp   1/1       Running   0          30m
my-nginx-9d5677d94-xh8q5   1/1       Running   0          30m

$ kubectl expose deployment my-nginx --port=80 --type=NodePort
service "my-nginx" exposed

$ kubectl get svc
NAME         TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)        AGE
kubernetes   ClusterIP   10.96.0.1        <none>        443/TCP        14h
my-nginx     NodePort    10.103.151.152   <none>        80:31277/TCP   5s

$ docker exec -it kube-master ifconfig eth0
eth0      Link encap:Ethernet  HWaddr 02:42:0a:c0:00:02  
          inet addr:10.192.0.2  Bcast:0.0.0.0  Mask:255.255.0.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
[...]

PR #55, opened 2017-12-11
Add MASTER_EXTRA_OPTS

ReSearchITEng:
Add MASTER_EXTRA_OPTS which will be used by docker run at on both: init and restore


ivan4th:
Thanks for the PR!

ReSearchITEng:
Thanks for the project :) 
Do you know any option to expose nodeport so we can access the apps deployed inside the cluster? (like  #43 ) ?

PR #58, opened 2017-12-19
add support for http proxies and custom certs

tallaxes:
This adds support for propagating HTTP proxies (and custom certificates), primarily for use behind corporate proxies. Should address https://github.com/Mirantis/kubeadm-dind-cluster/issues/52

In my current environment (Docker for Windows) the configured proxy setting are already propagated by the outer Docker/Moby to `kubeadm-dind-cluster` "node" containers (kube-master, kube-node-n) - including to anything executed there with `docker exec`; and further to the static pods configured by `kubeadm`, such as `apiserver`, etc. (`kubeadm` propagates all *_proxy variables it observers to `/etc/kubernetes/manifests`). The settings are _not_ propagated to the inner `dockerd` - because it is started via `systemd`, with  clean environment, - or to any other containers started with kubectl (dashboard, …). This PR takes care of propagating proxy settings into the `systemd` configuration of the inner `dockerd`. It does not compute proxy settings, just propagates them.

One can either set `DIND_PROPAGATE_HTTP_PROXY`, to use whatever proxy environment variables are already visible form within containers (useful if already propagated from configuration by the outer Docker), or explicitly set `DIND_HTTP_PROXY`, `DIND_HTTPS_PROXY`, `DIND_NO_PROXY`.

For custom certificate one can specify certificate URL in `DIND_CA_CERT_URL`, and it will be added to `/usr/local/share/ca-certificates`.

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -106,6 +106,12 @@ SKIP_SNAPSHOT="${SKIP_SNAPSHOT:-}"
 E2E_REPORT_DIR="${E2E_REPORT_DIR:-}"
 DIND_NO_PARALLEL_E2E="${DIND_NO_PARALLEL_E2E:-}"
 
+DIND_CA_CERT_URL="${DIND_CA_CERT_URL:-}"
+DIND_PROPAGATE_HTTP_PROXY="${DIND_PROPAGATE_HTTP_PROXY:-}"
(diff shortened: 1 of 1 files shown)


ivan4th:
Thanks for the PR!

PR #60, opened 2017-12-22
Allow KubeAdm config customizing and port forwarding.

pmichali:
Instead of hard coding the kubeadm.conf settings as part of the
dind-cluster.sh script, we'll use a template file and customize
it with the known settings needed based on the IP mode and CNI
mode. This allows users to add additional customization desired
(e.g. for istio), without altering scripting.

For minion nodes, allow port forwarding, using the EXTRA_PORTS
configuration setting.

As part of setting the static route to the NAT64 server for IPv6,
set the docker flag to remove the temp container used.


ivan4th:
There's a problem with this approach, namely that separate tmpl file is not compatible with the idea of having a self-contained `fixed/` scripts. So after `build/genfixed.sh` the `fixed/` scripts will be broken. Maybe we can somehow move the template into the image?

pmichali:
@ivan4th Can you elaborate on how the fixed scripts would be broken?

As I see it right now, the scripts currently depend on config.sh. This change would have them also depend on kubeadm.conf.tmpl, whose defaults should work for all current releases.

The thought of using this template, is that it would allow users to customized kubeadm.conf without modifying the dind-cluster.sh script. For example, for Istio, one could add the following to  apiServerExtraArgs:

     runtime-config: "admissionregistration.k8s.io/v1alpha1"

If we include the template in the image, won't it require an image rebuild, to customize the config?

pmichali:
@ivan4th Can you let me know your thoughts on this change set?  Also, has a 1.9 release been ulled yet for DinD?

stealthybox:
@pmichali the fixed v1.9 script is up, but the 1.9 dind image isn't published yet:  #67

image/wrapkubeadm
```diff
@@ -210,7 +213,7 @@ function dind::frob-proxy-configmap {
   #   max: 0
   #   maxPerCore: 0
   # masqueradeAll: true
-  local cluster_cidr="$(dind::cluster-cidr)"
+  local cluster_cidr=""
```
pmichali: Looks like I need to set this one too... Will push another change...

pmichali:
@ivan4th Please take a look at the change set, it now has four things:

- Pulled out kbeadm.conf so that it can be customized, if desired (though image rebuild needed).
- For IPv6, do not set cluster CIDR for kube-proxy to fix issue with kube-dns startup.
- Allow port forwarding to be specified for minion nodes.
- Remove temp container used to set up static routes.

Tested with IPv4 and IPv6 on bare-metal setup.

pmichali:
@ivan4th Updated version specific scripts, by running build/genfixed.sh and pushed up those changes. Please review.

PR #63, opened 2017-12-29
Fix pod default network for Calico CNI

### LINKED ISSUE #62: Pod network is not changed when using Calico CNI
There's a typo in bash scripts due to which `$POD_NETWORK_CIDR` is not changed to "192.168.0.0/16". I'm not quite sure why such CIDR adjustment is needed, but it does not work as expected.

**Steps to reproduce:**
```
SKIP_SNAPSHOT=true CNI_PLUGIN=calico-kdd ./dind-cluster-v1.8.sh up
```
and check Pods IPs

**Expected result:**
Pods have IPs from"192.168.0.0/16" network

**Actual result:**
Check Pod IPs, they are from "10.244.0.0/16" network

adidenko:
Fix typo in conditional and move CNI_PLUGIN variable definition a bit to make more sense (define first, then do conditional).

Closes #62

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -49,11 +49,11 @@ if [[ ! ${EMBEDDED_CONFIG:-} ]]; then
   source "${DIND_ROOT}/config.sh"
 fi
 
+CNI_PLUGIN="${CNI_PLUGIN:-bridge}"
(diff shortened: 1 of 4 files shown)


ivan4th:
Thanks for the PR!

PR #75, opened 2018-02-05
Enable mount propagation

lukaszo:
I wanted to make this optional but I don't see a clear way to do it. As long as hostpath is not used it shouldn't be noticed

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -746,6 +746,10 @@ EOF
   else
     init_args=(--pod-network-cidr="${POD_NETWORK_CIDR}")
   fi
+  # required when building from source
+  if [[ ${BUILD_KUBEADM} || ${BUILD_HYPERKUBE} ]]; then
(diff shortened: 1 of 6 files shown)


dind-cluster.sh
```diff
@@ -953,7 +954,8 @@ function dind::fix-mounts {
     if ((n > 0)); then
       node_name="kube-node-${n}"
     fi
-    docker exec "${node_name}" mount /run -o remount,rshared
+    docker exec "${node_name}" mount --make-shared /lib/modules/
+    docker exec "${node_name}" mount --make-shared /run
```
jellonek: Shouldn't that be `--make-rshared`?

lukaszo: In this case it shouldn't matter because kubelet is making them rshared when used in pods.

ivan4th:
The job for k8s master fails for some reason unrelated to this PR, will check it.

PR #81, opened 2018-02-13
In case of linuxkit/moby skip /lib/modules

lukaszo:
This should fix issue described here https://github.com/Mirantis/kubeadm-dind-cluster/issues/79#issuecomment-364492756

@mwpeterson I don't have a OSX to test it. Can you confirm that it fixes your issue?


mwpeterson:
Works perfectly!

ivan4th:
Yeah, works for me too, thanks a lot @lukaszo !

Issue #83, opened 2018-02-18
using_linuxkit causes dind-cluster-*.sh to fail on Docker for Windows 10


savamilovanovic:
After _using_linuxkit_ was introduced, it causes dind-cluster-*.sh to fail on Docker for Windows 10 with error **_mount: mount point /lib/modules/ does not exist_**:
```
sava@DellXPS:/mnt/c/k8s/kubeadm-dind-cluster/fixed$ ./dind-cluster-v1.8.sh up
* Making sure DIND image is up to date
* Removing container: c623f6231d2a
c623f6231d2a
* Removing container: 8cbd948ca416
8cbd948ca416
* Removing container: 9046fa7bd8cf
9046fa7bd8cf
* Restoring master container
* Restoring master container
* Restoring node container: 1
* Restoring node container: 2
* Starting DIND container: kube-node-2
* Starting DIND container: kube-node-1
* Starting DIND container: kube-master
* Node container restored: 1
* Node container restored: 2
* Master container restored
mount: mount point /lib/modules/ does not exist
```
Fix that worked for me was **_not_** to **use linuxkit** on Docker for Windows, by adding additional check `if ! docker info|grep -s '^Operating System: .*Docker for Windows' > /dev/null 2>&1 ; `  as shown below:
```
# In case of linuxkit / moby linux, -v will not work so we can't
# mount /lib/modules and /boot. Also we'll be using localhost
# to access the apiserver
using_linuxkit=
if ! docker info|grep -s '^Operating System: .*Docker for Windows' > /dev/null 2>&1 ; then
    if docker info|grep -s '^Kernel Version: .*-moby$' >/dev/null 2>&1 ||
         docker info|grep -s '^Kernel Version: .*-linuxkit-' > /dev/null 2>&1 ; then
        using_linuxkit=1
    fi
fi
```
Could this fix be worth adding onto main branch, please?

lukaszo:
Fixed in #84

PR #84, opened 2018-02-19
Proposed fix for issue #83: do not use linuxkit on Docker for Windows.

savamilovanovic:


ivan4th:
I don't have Windows installation to test it myself, but let's hope it'll work...

PR #85, opened 2018-02-20
Remove the --require-kubeconfig=true flag which is obsolete in 1.10

### LINKED ISSUE #82: kubelet never comes up on GCE with latest Kubernetes
When I run the gce-setup.sh script on the latest Kubernetes master, I see this error:
```
W0215 20:40:49.242] [init] Waiting for the kubelet to boot up the control plane as Static Pods from directory "/etc/kubernetes/manifests".
W0215 20:40:49.243] [init] This might take a minute or longer if the control plane images have to be pulled.
W0215 20:41:29.243] [kubelet-check] It seems like the kubelet isn't running or healthy.
W0215 20:41:29.243] [kubelet-check] The HTTP call equal to 'curl -sSL http://localhost:10255/healthz' failed with error: Get http://localhost:10255/healthz: dial tcp [::1]:10255: getsockopt: connection refused.
```
When I look at the kubelet journalctl log, I see this error:
```
Feb 16 01:29:27 kube-master hyperkube[2517]: F0216 01:29:27.361449    2517 server.go:142] unknown flag: --require-kubeconfig
```
The cause of this error seems to be this commit in kubernetes/kubernetes, which eliminated the deprecated "--require-kubeconfig" flag from kubelet configuration:
   [https://github.com/kubernetes/kubernetes/pull/58367](https://github.com/kubernetes/kubernetes/pull/58367)

Seems to me that this setting in k-d-c/image/kubelet.service needs this change for latest Kubernetes (although this would need to be a kubernetes-version-dependent change):
```
-Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf --require-kubeconfig=true"
+Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf"
```
One quick workaround is to back up to the version of Kubernetes right before the commit of k8s PR #58367, e.g.:
```
cd $GOPATH/src/k8s.io/kubernetes
git checkout ee062f9
```

cosmincojocar:
kubelet service does not start because of the `--require-kubeconfig` flag which is obsolete beginning  with version 1.10.

Fix #82

### DIFF
--- a/image/kubelet.service
+++ b/image/kubelet.service
@@ -3,7 +3,7 @@ Description=kubelet: The Kubernetes Node Agent
 Documentation=http://kubernetes.io/docs/
 
 [Service]
-Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf --require-kubeconfig=true"
+Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf"
 Environment="KUBELET_SYSTEM_PODS_ARGS=--pod-manifest-path=/etc/kubernetes/manifests --allow-privileged=true"
 Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin"
 Environment="KUBELET_DNS_ARGS=--cluster-dns=10.96.0.10 --cluster-domain=cluster.local"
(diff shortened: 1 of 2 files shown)


cosmincojocar:
@ivan4th Please could you have a look at this fix? Thanks

pigmej:
I like it.

@ivan4th can we merge that?

ivan4th:
Yeah, thanks, merging this one.

PR #93, opened 2018-03-16
Possibility to configure registry-mirrors and insecure-registries

pigmej:
It's needed for possibility to cache images for all registries
(sadly docker ignores insecure flag for registry-1.docker.io therefore separate registry mirror is required)

Best tested with: https://github.com/pigmej/docker-cache-all

Then you need to set env variables:
```
DIND_INSECURE_REGISTRIES="[\"0.0.0.0/0\"]"
DIND_REGISTRY_MIRROR="http://{YOUR_HOST_IP}:5000
DIND_HTTP_PROXY=http://{YOUR_HOST_IP:3128
```

then use dind normally

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -120,6 +120,9 @@ DIND_HTTP_PROXY="${DIND_HTTP_PROXY:-}"
 DIND_HTTPS_PROXY="${DIND_HTTPS_PROXY:-}"
 DIND_NO_PROXY="${DIND_NO_PROXY:-}"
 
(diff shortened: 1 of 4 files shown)


ivan4th:
Thanks!

abdennour:
What does it mean "YOUR_HOST_IP"? Private IP or external IP ?

pigmej:
@abdennour it's basically any IP of the host where you run k-d-c. You could set it to docker network ip assigned to your host. (docker0 iface if I recall correctly)

PR #102, opened 2018-03-29
change mount points for /sys/kernel/debug to shared to support Weave

jeffersonjhunt:
Weave Scope AddOn throws an error and wont gather data as a result. The error in question is /sys/kernel/debug isn't mounted shared. A quick change to the fix-mounts function allows Weave Scope to run as expected.


ivan4th:
Hi! Thanks for the PR. Please run `build/genfixed.sh` to update scripts in `fixed/`.

PR #107, opened 2018-04-06
Use separate pod networks, instead of nesting under dindnet.

### LINKED ISSUE #106: Need to separate pod networks from dind network
Currently ,the pod networks are subnets of the dind network. During E2E testing, we are seeing about eight tests failing, with that setup (they work fine on KubeAdm clusters where the pod networks are independent of the management network).

In investigating these failures they are where services are spun up on each minion node, and then the test tries to communicate with the service on it's port. The packets are getting to the pod on the same node as the request, but not to the pod on the other minion.

A workaround was found, by turning on promiscuous mode on the dind0 interface on all nodes, but we felt that this was not a good solution to employ for the cluster.

By changing the pod networks, so that they are subnets of a network, independent of the dind network, and creating static routes on the nodes, everything worked fine.

Will be pushing up a PR to implement this solution.

pmichali:
There are several parts to this change. First, we'll use the existing
POD_NETWORK_CIDR variable to indicate the cluster for pod networks.
Each node will have a subnet in that cluster, that uses the node's ID
as the discriminator. For example, 10.244.0.0/16 (the default), will
define 10.244.1.0/24 for kube-master, 10.244.2.0/24 for the first
minion, etc.

Second, the POD_NETWORK_CIDR will be broken into a prefix and a size,
that is passed to dindnet, to properly provision the CNI config file
for bridge plugin. For IPv6, the code will pad the prefix with zeros,
so that it is fully qualified and then append the node's ID, giving a
proper prefix that can be used to create the GW IP and network address.

There are a few cases for the forming of the prefix in IPv6. The simple
case is where the POD_NETWORK_CIDR of "fd00:40::/72" is converted into a
prefix of "fd00:40:0:0:" and the node ID (e.g. 2) is added, to give a prefix
of "fd00:40:0:0:2". On the node, this would give a pod network value of
"fd00:40:0:0:2::/80".

Another case is when the user specifies a fully qualified POD_NETWORK_CIDR
of "fd00:10:20:30:4000::/72". This will be truncated to the prefix value of
"fd00:10:20:30:40", and then with the node ID added, we get pod network of
"fd00:10:20:30:4002::/80" on a node with ID 2.

The last case, is with a POD_NETWORK_CIDR of "fd00:40::/64", where the
prefix will be "fd00:40:0:0:", and the node ID will be placed in the upper
byte of the next word of the prefix. On a node with ID 2, this gives a pod
network of "fd00:40:0:0:0200::/72".

Third, the CNI config uses ipMasq (for outside access), isDefaultGateway
(to give dind0 interface an IP), and no longer specifies the minimum
and maximum host IP for the pod network, default route, or route to DNS64
network via the NAT64 (as that is added to the eth0 interface).

Fourth, for IPv4, we'll assume that the POD_NETWORK_CIDR has a /16 size
and pod networks are /24.  For IPv6, pod networks will be 8 bits smaller
than the cluster size as well.

Fifth, because the pod networks are no longer nested inside of the dindnet,
static routes are added on each node, so that pods can access pods on
other nodes.

Sixth, for E2E tests we need to have hairpin mode enabled int he CNI config
file. Created USE_HAIRPIN environment variable, which will default to false
for IPv4 (for backwards compatibility, as there is some indication of an
issue when using hairpin mode and Virtlet Networking), and will default to
true for IPv6.

Fixes #106

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -59,11 +59,6 @@ if [[ ! ${EMBEDDED_CONFIG:-} ]]; then
 fi
 
 CNI_PLUGIN="${CNI_PLUGIN:-bridge}"
-DEFAULT_POD_NETWORK_CIDR="10.244.0.0/16"
-if [[ ${CNI_PLUGIN} = "calico" || ${CNI_PLUGIN} = "calico-kdd" ]]; then
-  DEFAULT_POD_NETWORK_CIDR="192.168.0.0/16"
-fi
-POD_NETWORK_CIDR="${POD_NETWORK_CIDR:-${DEFAULT_POD_NETWORK_CIDR}}"
 ETCD_HOST="${ETCD_HOST:-127.0.0.1}"
 GCE_HOSTED="${GCE_HOSTED:-false}"
 if [[ ${IP_MODE} = "ipv6" ]]; then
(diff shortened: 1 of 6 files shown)


dind-cluster.sh
```diff
@@ -852,6 +892,41 @@ function dind::kill-failed-pods {
...
+	# Assuming pod subnets will all be /24
+        dest="${POD_NET_PREFIX}${id}.0/24"
+        gw=`docker exec $dest_node ifconfig eth0 | grep "inet addr" | cut -f 2 -d: | cut -f 1 -d' '`
+      else
+	instance=$(printf "%02x" ${id})
+	if [[ $((${POD_NET_SIZE} % 16)) -ne 0 ]]; then
+	  instance+="00" # Move node ID to upper byte
+	fi
+	dest="${POD_NET_PREFIX}${instance}::/${POD_NET_SIZE}"
+        gw=`docker exec $dest_node ifconfig eth0 | grep "inet6" | grep -i global | cut -b 23- | cut -f 1 -d/`
```
pmichali: Per our discussion, will change this to use awk, instead of cut -b, so that is not dependent on exact output, in case things change in future. Thanks!

pmichali:
Just pushed up changes based on review comments, and then additional changes to clean up style used on variable naming.

pmichali:
@ivan4th PTAL at the latest version.

ivan4th:
Thanks! It's ok now.

pmichali:
Thanks for the review/approval @ivan4th

Issue #127, opened 2018-05-21
IPv6: kube-dns restarts during E2E tests


pmichali:
As mentioned in https://github.com/kubernetes/kubernetes/issues/63922, kube-dns (and coredns) was restarting during E2E tests, when running on an IPv6 cluster. These restarts would sometimes cause test failures, as kube-dns would not be ready, causing a pre-test timeout/failure.

Found that increasing the MTU used, would resolve the issue, which applies to any environment (k-d-c or bare-metal).

Working on a solution to use a larger MTU size.

pmichali:
On bare-metal, tried using MTU 9000 for interfaces used for the pod and management networks. Ran E2E tests for IPv6, and they all passed (6 times) without any kube-dns restarts.

Under kubeadm-dind-cluster environment, which is going to be used to run IPv6 E2E tests for CI, this MTU change, in and of itself, did NOT resolve the issue. If I turn off hairpin mode (which breaks the node-service UDP and HTTP tests as they are using a pod that refers to service on the same pod), then it seems that the  E2E runs are completing w/o kube-dns restarts.

Need to get a better understanding of what is causing kube-dns to fail to respond to liveness/readiness probes, and therefore getting restarted by kubelet.

pmichali:
Found root cause, and it is not MTU. Turns out that MAC of dind0 bridge is changing during the test runs. Found that if MAC is not set explicitly, then it will get the lowest numbered MAC of all enslaved interfaces. It appears that, if an new enslaved interface is added to the bridge, with a lower MAC, the MAC of the bridge will get the new value. This messes with the TCP communication to kube-dns running on the pod associated with the bridge.

Solution is to explicitly assign a MAC to the dind0 bridge. WIll do that by creating a veth pair, adding one to the dind0 bridge and then setting the bridge MAC to match the veth MAC. This will force it to stay the same.

PR #135, opened 2018-06-01
Force MAC on pod network bridge

pmichali:
In running E2E tests, it was observed that during the testing, the MAC
on the dind0 bridge was changing. This was causing, in some cases, the
kube-dns pod to fail health checks (lose TCP communication on the bridge
for that pod).

Research shows that, if a MAC is not assigned to a virtual bridge, then
the bridge will get the lowest MAC of all enslaved interfaces (dynamically).
This implies that the MAC could also change, and in this case, if a test
pod has a lower MAC than kube-dns (which is the first pod added to the
dind0 bridge at startup), then the MAC of the bridge would change.

This does not appear to be a problem for IPv4, where the bridge has a MAC
address assigned by the CNI plugin.

With the latest CNI plugin, v0.7.1, the plugin will set the MAC for
bridges created. It will adopt existing bridges (like dind0), but will
not ensure they have a MAC.

I did try changing the sequencing of the dindnet script to first create
the CNI config, and then create the bridge, but it appears that CNI does
not get involved in the bridge creation at that time (it does later).

As a solution, this commit will create a dummy veth pair, add one of the
interfaces into the bridge, and then set the bridge's MAC to the MAC of
the interface (reading up on this indicates that we must use a MAC of an
enslaved interface).

Now, dind0 has a MAC address, and since it was explicitly set, there is no
dynamic override of the address. During E2E tests, kube-dns health check
responses make it back to the kubelet, and it no longer gets restarted.

NOTE: This commit uses CNI v0.7.1, which is found in a different repo
location. We need this newer version, as v0.6 does not have IPv6 support,
and caused other issues (like IP6tables not being set up correctly).

Fixes Mirantis issue: #127 
Fixes Kubernetes issue: #63922

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -896,14 +896,14 @@ function dind::create-static-routes-for-bridge {
       if [[ ${IP_MODE} = "ipv4" ]]; then
 	# Assuming pod subnets will all be /24
         dest="${POD_NET_PREFIX}${id}.0/24"
-        gw=`docker exec ${dest_node} ifconfig eth0 | grep "inet addr" | cut -f 2 -d: | cut -f 1 -d' '`
+        gw=`docker exec ${dest_node} ip addr show eth0 | grep -w inet | awk '{ print $2 }' | sed 's,/.*,,'`
       else
 	instance=$(printf "%02x" ${id})
 	if [[ $((${POD_NET_SIZE} % 16)) -ne 0 ]]; then
 	  instance+="00" # Move node ID to upper byte
 	fi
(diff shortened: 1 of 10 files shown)


pmichali:
Tested with IPv4 and IPv6 clusters, using k-d-c localling and via GCE.

ivan4th:
Would you please rebase the PR on top of `master`? 04ce596 commit seems to be out of place here

ivan4th:
Closing, replaced by #144

Issue #147, opened 2018-06-08
IPv6: Allow AAAA records to be used...


pmichali:
As currently configured, the IPv6 cluster created will always use A records for DNS lookups. As a result, accessing external sites that support IPv6 addresses, DNS will discard the (IPv6) AAAA record and use the (IPv4) A record, creating a synthesized IPv6 address. All access to that site will then use NAT64.

A performance enhancement, for environments that have IPv6 access to the outside, we can allow AAAA records to be used and then DNS will return the IPv6 address. Access to external sites with IPv6 addresses do not have to use NAT64.

I'm working on a change to provide a know to allow this mode to be enabled.

PR #148, opened 2018-06-08
IPv6: Allow AAAA record use

pmichali:
By default, DNS is set up to ignore (IPv6) AAAA records. As a result,
whenever a remote site is accessed, only the (IPv4) A record is used,
and a synthesized IPv6 address formed. This means that NAT64 will be
used for all external accesses, even when a site supports IPv6.

This is fine for environments where the cluster only has IPv4 access,
like GCE currently. However, if the cluster will have IPv6 access too,
the configuration can be modified to use AAAA records in DNS lookups,
and then remote sites with IPv6 addresses can be directly accessed
(without using NAT64).

This change provides a knob to allow enabling use of AAAA records. In
addition to modifying the DNS configuration, ip6tables rules are
created to perform SNAT for packets coming from the container nodes
and to forward external packets to the kubeadm-dind-net to get to the
pods.

The boolean environment variable, ALLOW_AAAA_USE, is defined for this
capability. The default is NOT to allow use of AAAA records, for
backwards compatibility.

Fixes issue #147

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -60,7 +60,8 @@ fi
 
 CNI_PLUGIN="${CNI_PLUGIN:-bridge}"
 ETCD_HOST="${ETCD_HOST:-127.0.0.1}"
-GCE_HOSTED="${GCE_HOSTED:-false}"
+GCE_HOSTED="${GCE_HOSTED:-}"
+DIND_ALLOW_AAAA_USE="${DIND_ALLOW_AAAA_USE:-}"  # Default is to use DNS64 always for IPv6 mode
 if [[ ${IP_MODE} = "ipv6" ]]; then
     DIND_SUBNET="${DIND_SUBNET:-fd00:10::}"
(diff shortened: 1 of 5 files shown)


dind-cluster.sh
```diff
@@ -76,6 +76,7 @@ if [[ ${IP_MODE} = "ipv6" ]]; then
     dns_server="${dind_ip_base}100"
     DEFAULT_POD_NETWORK_CIDR="fd00:10:20::/72"
     USE_HAIRPIN="${USE_HAIRPIN:-true}"  # defaults on
+    ALLOW_AAAA_USE="${ALLOW_AAAA_USE:-false}"  # Default is to use DNS64 always
```
rpothier: Would it make sense going forward that we prefix new env variables with DIND_?
I know only some have the prefix at this point, (DIND_SUBNET, DIND_IMAGE ...) but it makes it easier to review the env variables.

pmichali: Sure, will update.

PR #149, opened 2018-06-18
Make feature-gates and any extra args customizable

everpeace:
Hi!  Thank you guys for maintaining this awesome projects!

I would like to improve customizability of dind-cluster on feature-gates and extra args because we wanted to test pod preemption behavior on dind-cluster.
 
Thus, I added environment variable to customize dind cluster:
- `FEATURE_GATES` which is passed to `kube-apiserver`, `kube-controller-manager` and `kube-scheduler`
  - default value is `MountPropagation=true`
  - `none` supported for not setting any feature gates
- `KUBELET_FEATURE_GATES` for `kubelet`
  - default value is `MountPropagation=true,DynamicKubeletConfig=true`
  - `none` supported for not setting any feature gates
- `APISERVER_underscoreed_option_name` for arbitrary extra args for `kube-apiserver` (e.g. `APISERVER_admission_control=xxx,yyy` will be converted to `--admission-control=xxx.yyy`
  - `SCHEDULER_xx_yy` and `CONTROLLER_MANAGER_xx_yy` are similar

For example, if users wanted to turn on priority and pod preemption, they can do in `config.sh`

```
FEATURE_GATES=MountPropagation=true,PodPriority=true
KUBELET_FEATURE_GATES=MountPropagation=true,DynamicKubeletConfig=true,PodPriority=true
APISERVER_admission_control=NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota,Priority
APISERVER_runtime_config=scheduling.k8s.io/v1alpha1=true
SCHEDULER_v=5  # this is useful for debugging preemption

# you need to build local image to try this branch
# DIND_IMAGE=mirantis/kubeadm-dind-cluster:local
```


everpeace:
/assign @ivan4th 

Could you review this PR?  It's my first PR.  So, I would be very glad if you gave some advice on this PR being merged Thank you in advance.

dind-cluster.sh
```diff
@@ -675,7 +677,15 @@ function dind::kubeadm {
   status=0
   # See image/bare/wrapkubeadm.
   # Capturing output is necessary to grab flags for 'kubeadm join'
-  if ! docker exec "${container_id}" /usr/local/bin/wrapkubeadm "$@" 2>&1 | tee /dev/fd/2; then
+  kubelet_feature_gates=""
+  if [ "z${KUBELET_FEATURE_GATES}" != "z" ]; then
```
ivan4th: In bash, it's better to use just `if [[ ${KUBELET_FEATURE_GATES} ]]; then`

dind-cluster.sh
```diff
@@ -675,7 +677,15 @@ function dind::kubeadm {
...
+    # delete 'MountPropergation=true, DynamicKubeletConfig=true' because it will be prepended in wrapkubeadm
```
ivan4th: Maybe we should make `KUBELET_FEATURE_GATES` default to `MountPropagation=true,DynamicKubeletConfig=true` and modify `wrapkubeadm` to use `KUBELET_FEATURE_GATES` value for the option value as is?
We can also support `KUBELET_FEATURE_GATES=none` that  would stand for not
adding any feature gates for kubelet at all.

dind-cluster.sh
```diff
@@ -675,7 +677,15 @@ function dind::kubeadm {
...
+    KUBELET_FEATURE_GATES=$(echo $KUBELET_FEATURE_GATES | sed -e 's/MountPropagation=true,\{0,\}//g' -e 's/,\+/,/g' -e 's/,$//g')
```
ivan4th: Is `,\{0,\}` the same as `,*` or am I missing something?

dind-cluster.sh
```diff
@@ -815,6 +825,31 @@ function dind::init {
     feature_gates="{CoreDNS: false}"
   fi
 
+  component_feature_gates="feature-gates: \\\"MountPropagation=true\\\""
+  if [ "z${FEATURE_GATES}" != "z" ]; then
```
ivan4th: Same as above, `if [[ ${FEATURE_GATES ]]; then`

config.sh
```diff
@@ -61,3 +61,22 @@ CNI_PLUGIN="${CNI_PLUGIN:-bridge}"
...
+# Kubelet Feature Gates
+# KUBELET_FEATURE_GATES=""
+
+# You can configure extra component args for kube-apiservers
+# APISERVER_underscored_option_name will be converted --hyphenated-option-name
+# e.g. APISERVER_admission_control=xxx,yyy -> --admission-control=xxx,yyy
+# APISERVER_xxx_yyy=zzz
+
+# Extra component args for kube-controller-manager
+# CONTROLLERMANAGER_underscored_option_name will be converted --hyphenated-option-name
```
ivan4th: Maybe we should call it `CONTROLLER_MANAGER` instead of `CONTROLLERMANAGER`?

image/wrapkubeadm
```diff
@@ -320,7 +316,7 @@ function dind::frob-cluster {
 
 kubelet_dind_args=(--fail-swap-on=false
                    --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf
-                   --feature-gates=DynamicKubeletConfig=true,MountPropagation=true)
+                   --feature-gates=DynamicKubeletConfig=true,MountPropagation=true${KUBELET_EXTRA_FEATURE_GATES:-})
```
ivan4th: As noted above, we probably should just use `KUBELET_EXTRA_FEATURE_GATES` as-is, with a special case of `none` which stands for not adding `--feature-gates` flag to the kubelect command line at all.

dind-cluster.sh
ivan4th: I.e. set the default like this `KUBELET_FEATURE_GATES="${KUBELET_FEATURE_GATES:-ountPropagation=true,DynamicKubeletConfig=true}" and also check for `none` value in wrapkubeadm.

everpeace:
@ivan4th Thank you for your review!  I fixed a place to set default values for feature gates into `dind-cluster.sh`.  Please take a look again 🙇

ivan4th:
Thanks!
/approve
/lgtm

PR #150, opened 2018-06-18
Add an extra space in sha1sum to work with alpine's sha1sum

ghost:
I'm running kdc in `docker:dind`.  v1.9 works but v1.10 gets hung up due to https://github.com/kubernetes/kubernetes/issues/61277

This PR is to fix sha1sum panicking in `docker:dind` because `sha1sum` is strict about the spacing of its input.

https://stackoverflow.com/a/37499133

Note that sha1sum expects two spaces between hash and filename.


darkyat (approved):

ivan4th:
Thanks for the PR!
/lgtm

Issue #158, opened 2018-07-10
Add support for IPVS CNI plugin


pmichali:
IPVS has gone GA in 1.11 and there is a push to make this replace kube-proxy.  Currently, it only supports IPv4, but work is in progress for IPv6.

I'll push up a PR for this, soon.

PR #159, opened 2018-07-10
Support for IPVS CNI Plugin

pmichali:
Added support for the plugin for IPv4 (does not support IPv6 yet). Can
be used by setting CNI_PLUGIN to "kube-router". As part of the setup,
the kube-proxy daemonset is deleted.

A few warnings occur during up/down, but appear to be benign (e.g.
kube-proxy warns to use a config file, no subnet to delete iptable
rules for IPVS).

Updated the "fixed" version scripts.

Fixes: #158

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -1029,8 +1029,12 @@ function dind::remove_external_access_on_host {
 }
 
 function dind::wait-for-ready {
-  dind::step "Waiting for kube-proxy and the nodes"
-  local proxy_ready
+  local app="kube-proxy"
+  if [[ ${CNI_PLUGIN} = "kube-router" ]]; then
+    app=kube-router
+  fi
+  dind::step "Waiting for ${app} and the nodes"
+  local app_ready
   local nodes_ready
   local n=3
   local ntries=200
@@ -1041,12 +1045,12 @@ function dind::wait-for-ready {
     else
       nodes_ready=y
     fi
-    if dind::component-ready k8s-app=kube-proxy; then
-      proxy_ready=y
+    if dind::component-ready k8s-app=${app}; then
+      app_ready=y
     else
-      proxy_ready=
+      app_ready=
     fi
-    if [[ ${nodes_ready} && ${proxy_ready} ]]; then
+    if [[ ${nodes_ready} && ${app_ready} ]]; then
       if ((--n == 0)); then
         echo "[done]" >&2
         break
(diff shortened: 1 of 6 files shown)


pmichali:
@ivan4th Any thoughts on the test_src_master failure. It seems to do kubeadm init and shows a message (doesn't say it is an error) and then gets stuck in a kubeadm reset loop.

```
your configuration file uses an old API spec: "kubeadm.k8s.io/v1alpha1". Please use kubeadm v1.11 instead and run 'kubeadm config migrate --old-config old.yaml --new-config new.yaml', which will write the new, similar spec using a newer API version.
```

It seems like, with V1.11, the API version should be v1alpha2.
Does master need branch need to be updated?

rpothier:
Can we just run kube-proxy in IPVS mode, instead of pulling in kube-router?
i.e. update kubeadm.conf with
kubeProxy:
  config:
    mode: ipvs

pmichali:
@rpothier Never tried that, and it's not listed as a way to setup networking using KubeAdm (whereas using kube-router is listed). Do you know if doing that will provide BGP (which is what kube-router does)? If not, then I'm not sure how it handles routing between nodes.

With kube-router, it uses bride CNI plugin, only for IPAM, and no hairpin or static routes need to be setup. In DinD, the YAML for kube-router sets up the CNI config.

pmichali:
Will give kube-proxy with IPVS to see how that works.

pmichali:
Clarification: Probably should have been clearer on this PR description, but partly it is an issue with the ambiguity of kube-router. Although called a CNI plugin, kube-router is really a networking component, which is really a replacement for kube-proxy, plus more. It has both the advantage of IPVS performance and scale, and the added functionality of BGP for dealing with the routing.

It really is not a CNI plugin, per se, and uses the bridge plugin, with some different settings. For k-d-c, I defined a different CNI_PLUGIN, so that it doesn't do the bridge plugin setup in dindnet script.  Like calico and others, kube-router has a YAML for setup. One thing it does in this YAML, is setup the bridge CNI config file, so functionality in dindnet is not needed (hence the separate CNI_PLUGIN type).

So, I'd like to make this available, as a CNI_PLUGIN "kube-router".  I'm peeking into running kube-proxy in IPVS mode, as @rpothier mentioned, but, if that is another viable method, I'd like to provide that as a separate option for k-d-c, and not as an alternative for kube-router, because kube-router does much more (BGP) and makes the setup even simpler (and works with external BGP,...).

IOW, I'd like to move this PR forward, independent of kube-proxy in IPVS mode.

pmichali:
Talked to @ivan4th and it looks like for master branch Kubernetes, the k-d-c image needs updating. Once that is done, I can rebase this PR to pickup the changes.

rpothier:
@pmichali Sounds good to me.

PR #161, opened 2018-07-11
Support for parallel clusters

mariantalla:
This PR enables dind-cluster to create and run multiple dind clusters at the same time.

# Some use cases:
1. Use dind to create test clusters
Functionality added by this PR would allow consuming tests to run in parallel. The [testing framework]() is an example.
1. Emulate a production setup, where multiple isolated clusters are commonly running

# Implementation Considerations
1. This PR changes the way that the [default container label](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/019854d91a3617c4e80ee438ff8424c7845ba38b/image/Dockerfile#L38) is used.
More specifically:
- The Dockerfile still specifies a label (image label). This means that any external systems that were depending on that label to discover containers created by `dind-cluster.sh` will continue to discover them. Note: the label will now apply to all containers of all clusters that the script creates.
- The image label is not used by `dind-cluster.sh` anymore. Instead, a [different default label](https://github.com/pivotal-k8s/kubeadm-dind-cluster/blob/5a9e45b745153d38c2ac6dac91a5482f9f959d68/dind-cluster.sh#L178-L179) (can be changed per cluster via environment variable) is used. This is to enable discovery of multiple clusters, and ensure that only the intended containers and resources are targeted by `up`/`down`/`clean` etc.

# UX Considerations
## - Explicit context setting
Previously a default context was created by `dind-cluster.sh` for the cluster; and used by kubectl. With this PR, a new context is created per cluster and then is explicitly set as part of the kubectl call when targeting an apiserver. This means that users also need to explicitly specify the context when they want to connect to a cluster.
## - Naming convention for docker resources
We originally proposed changing the convention from `kube-master`/`kube-node`/`kube-net` etc. to a suffixed version of the names. The suffix is a sha1 string. The intention here is to keep the resource names unique, to satisfy docker constraints. Any external consumers of the clusters created by `dind-cluster.sh` will have to update any references to the non-suffixed resource names.

Eventually the UX is: when creating a cluster using the default label (no `DIND_LABEL` set), the naming of the resources stays as it was pre-this-PR, i.e. `kube-master`/`kube-node`/`kube-net` etc.; this covers use cases mentioned in [this comment](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/161#issuecomment-404889495).
When a custom label is set through `DIND_LABEL`, resource names will be suffixed with the label's sha, as described above.

cc @hoegaarden

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -266,22 +266,24 @@ function dind::prepare-sys-mounts {
     fi
     return 0
   fi
-  if ! dind::volume-exists kubeadm-dind-sys; then
+  local dind_sys_vol_name
+  dind_sys_vol_name="kubeadm-dind-sys-$( dind::sha1 "$DIND_LABEL" )"
+  if ! dind::volume-exists "$dind_sys_vol_name"; then
     dind::step "Saving a copy of docker host's /lib/modules"
-    dind::create-volume kubeadm-dind-sys
+    dind::create-volume "$dind_sys_vol_name"
     # Use a dirty nsenter trick to fool Docker on Mac and grab system
     # /lib/modules into sys.tar file on kubeadm-dind-sys volume.
     local nsenter="nsenter --mount=/proc/1/ns/mnt --"
     docker run \
            --rm \
            --privileged \
-           -v kubeadm-dind-sys:/dest \
+           -v "$dind_sys_vol_name":/dest \
            --pid=host \
            "${busybox_image}" \
            /bin/sh -c \
            "if ${nsenter} test -d /lib/modules; then ${nsenter} tar -C / -c lib/modules >/dest/sys.tar; fi"
   fi
-  sys_volume_args=(-v kubeadm-dind-sys:/dind-sys)
+  sys_volume_args=(-v "$dind_sys_vol_name":/dind-sys)
 }
 
 tmp_containers=()
@@ -729,9 +731,11 @@ function dind::set-master-opts {
   master_opts=()
   if [[ ${BUILD_KUBEADM} || ${BUILD_HYPERKUBE} ]]; then
     # share binaries pulled from the build container between nodes
-    dind::ensure-volume "dind-k8s-binaries"
+    local dind_k8s_bin_vol_name
+    dind_k8s_bin_vol_name="dind-k8s-binaries-$(dind::sha1 "$DIND_LABEL")"
+    dind::ensure-volume "${dind_k8s_bin_vol_name}"
     dind::set-build-volume-args
-    master_opts+=("${build_volume_args[@]}" -v dind-k8s-binaries:/k8s)
+    master_opts+=("${build_volume_args[@]}" -v "${dind_k8s_bin_vol_name}:/k8s")
     local -a bins
     if [[ ${BUILD_KUBEADM} ]]; then
       master_opts+=(-e KUBEADM_SOURCE=build://)
@@ -896,7 +900,7 @@ function dind::create-node-container {
   local node_ip="${dind_ip_base}$((next_node_index + 2))"
   local -a opts
   if [[ ${BUILD_KUBEADM} || ${BUILD_HYPERKUBE} ]]; then
-    opts+=(-v dind-k8s-binaries:/k8s)
+    opts+=(-v "dind-k8s-binaries-$(dind::sha1 "$DIND_LABEL")":/k8s)
     if [[ ${BUILD_KUBEADM} ]]; then
       opts+=(-e KUBEADM_SOURCE=build://)
     fi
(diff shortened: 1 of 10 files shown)


hoegaarden:
Is there a way to rerun one (or all) circleci tests? We think this is a flake and would love to just rerun instead of pushing random commits.

hoegaarden:
Please take a look and let us know what you think about that feature -- esp. in regards to the UX changes (no more default context, all things suffixed with a lovely hash, ...)
Thanks a lot, Hannes & Maria

/cc @timothysc

ivan4th:
Thanks for the PR. It'll be an important addition to kubeadm-dind-cluster. One thing though: how hard is to preserve the old behaviour with predictable node container names? E.g. when DIND_LABEL is not set. Thing is, I expect that in many situations people have scripts in CI etc that use things like `docker exec kube-node-1 ...`, e.g. we have this in [Virtlet](https://github.com/Mirantis/virtlet). I understand that this may make the code more complicated and less elegant, but still I'm afraid there'll be complaints from some users otherwise.

dind-cluster.sh
```diff
@@ -1242,20 +1271,45 @@ function dind::down {
...
+  nameRE='^kubeadm-dind-.*-'"$(dind::sha1 "$DIND_LABEL")"
+  docker volume ls -q | (grep "$nameRE" || true) | while read volume_id; do
     dind::step "Removing volume:" "${volume_id}"
     docker volume rm "${volume_id}"
   done
 }
 
+function dind::sha1 {
+  # shellcheck disable=SC2046
+  set -- $( echo "$@" | sha1sum )
```
ivan4th: perhaps `echo -n "$@" | sha1sum` as extra "\n" produced by the echo being included in the hash is a bit non-obvious.

ivan4th:
As of the flakes, I'll try to see if it's possible to make CircleCI allow non-owners to retrigger the jobs or if it's somehow doable using k8s-ci-robot.

hoegaarden:
@ivan4th 

> One thing though: how hard is to preserve the old behaviour with predictable node container names? E.g. when DIND_LABEL is not set.

Are you suggesting to have clusters with no `DIND_LABEL` set to have container names like `kube-master` and only clusters with `DIND_LABEL` set to have container names like `kube-master-<hash>`? I think that would be rather easy, given we have extracted the naming stuff for the containers in their own functions.

ivan4th:
@hoegaarden 
> Are you suggesting to have clusters with no DIND_LABEL set to have container names like kube-master and only clusters with DIND_LABEL set to have container names like kube-master-<hash>? I think that would be rather easy, given we have extracted the naming stuff for the containers in their own functions.

Yes, that's what I mean.

Concerning retriggering the jobs: it's a bit clumsy, but doable, just checked from another GH account (I was not sure if people other than owners can rerun the jobs). You need to find the workflow (CircleCI's pipeline) for the PR. You can just use URL like https://circleci.com/gh/kubernetes-sigs/kubeadm-dind-cluster/tree/pull/161 and click on `build-test-push` (the name of the workflow) in any of the job rows, or click Details in the list of PR's checks, then find `build-test-push` on the page (using Cmd-F / Ctrl-F 
for example) and click it. On the workflow page, there's Rerun button (again, use Cmd-F / Ctrl-F if CircleCI UI confuses you). It has dropdown menu, so you can either restart all of the jobs in the workflow or only restart the failed ones. Note that in some cases the restart doesn't happen immediately (maybe depending on the load of CircleCI servers).

Sorry for the inconvenience, I was using Travis CI before but it turned out that there's no good way to use ssh for debugging the jobs with Travis (it'll leak your secrets like credentials for the image repo), so I switched to CircleCI for the most of the tests except a few that are there to make sure KDC is still compatible with Travis CI. In Travis, the restart button is clearly visible and no workflows / pipelines are in use atm.

ivan4th:
Sorry about the conflict, a PR was merged that fixes k8s master build.

mariantalla:
@ivan4th - no worries we'll rebase.

We had a question about running `test.sh` test cases in CI. The file doesn't appear in the CIrcleCI config. Some test cases run in Travis (`TEST_CASE=1.8`, `TEST_CASE=1.9`, `TEST_CASE=1.10`). Do we go ahead and add any new ones there? Is there a different pattern we should follow?

cc @hoegaarden

ivan4th:
@mariantalla intended usage of test.sh is (1) running tests locally (2) running a subset of tests on Travis CI to make sure k-d-c still works there. The idea is to run tests for each supported stable k8s version. I've forgot to include 1.11 there after adding it, thanks for noticing, will fix it a bit later.

mariantalla:
@ivan4th Makes sense. In that case we were planning to move/add the tests that cover this PR to the jobs that run on CircleCI - so that they keep running on future commits. Any objections?

hoegaarden:
@ivan4th We added some tests for multiple parallel clusters on CIrcleCI (`test_multiple_clusters_$VER`). From our perspective this PR is ready to review. Please let us know what you think about it. Thanks!

ivan4th:
Thanks for the PR! Merging.

PR #166, opened 2018-07-13
Find free subnet and port

hoegaarden:
This eases the usage of multiple parallel clusters. Users now do not need to specify the network and local apiserver port manually -- it will be automatically found.

#### APIServer port is completely random

We leave it to docker to decide which port should be used for the port forwarding. This means that there is no default anymore (previous to this PR the default was `8080`), so also "default" clusters (without `DIND_LABEL` set) will use a random port.

#### Subnet for docker network is random

The subnet for the cluster will be randomly chosen. Previous to this PR the default subnet was `10.192.0.0`. Now, we ask docker on which networks are already allocated and try to use a subnet right after the "highest" allocated IP. Only if docker has no subnets allocated yet, we use `10.192.0.0`.
The whole automatism also works when different subnet sizes are in play.

One caveat: This mechanism is racy. Something else might claim a subnet between the time where we query docker for its networks and we actually create the network.

#### Port forwarder started by `dind-cluster.sh`

As the APIServer port it now completely random, it is hard to set up the port forwarding (in case of a remote docker host) up front. So we added a new config option to `dind-cluster`: `DIND_PORT_FORWARDER`
- Users can set this variable to a path to a executable
- When we now the random APIServer port, we call out to that executable
- We pass the random port as the first argument to the executable
- When `DIND_PORT_FORWARDER` is not set / empty we ignore it and don't do any external call

#### Configuration options as before

The variables `DIND_SUBNET`, `DIND_SUBNET_SIZE` & `APISERVER_PORT` will still be used if set.

#### New subcommand `apiserver-port`

Calling `./dind-cluster.sh apiserver-port` can be used to query which port the APIServer of a cluster is listening on.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -192,8 +192,8 @@ test_multiple_clusters: &test_multiple_clusters
         name: Bring up multiple clusters in parallel
         no_output_timeout: 35m
         command: |
-          build/portforward.sh -wait 8080
-          build/portforward.sh -wait 8082
+          build/portforward.sh -wait 8080  # by default, first cluster binds to 8080
+          build/portforward.sh -wait 8081  #       ... and second cluster binds to 8081
 
           set -eu
           set -o pipefail
@@ -205,7 +205,7 @@ test_multiple_clusters: &test_multiple_clusters
             local d="./dind-cluster.sh"
 
             "${d}" up
-            APISERVER_PORT=8082 DIND_SUBNET='10.199.0.0' DIND_LABEL="$customLabel" "${d}" up
+            DIND_LABEL="$customLabel" "${d}" up
 
             # masters
             test "$(countContainersWithExactName "kube-master")" -eq 1 || {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,19 +31,21 @@ setup: &setup
 
 display_cluster_state: &display_cluster_state
   command: |
-    build/portforward.sh -wait 8080
     kubectl="kubectl"
     if [[ ${K8S_SRC:-} ]]; then
       cd kubernetes
       kubectl="cluster/kubectl.sh"
     fi
-    "${kubectl}" version
-    "${kubectl}" get all --all-namespaces -o wide
-    "${kubectl}" get nodes
+
+    apiserver_port="$( ./dind-cluster.sh apiserver-port )"
+    build/portforward.sh -wait "$apiserver_port"
+
+    "${kubectl}" "--server=:${apiserver_port}" version
+    "${kubectl}" "--server=:${apiserver_port}" get all --all-namespaces -o wide
+    "${kubectl}" "--server=:${apiserver_port}" get nodes
 
 dump_cluster: &dump_cluster
   command: |
-    build/portforward.sh -wait 8080
     mkdir -p /tmp/cluster_state
     out="/tmp/cluster_state/dump-1.gz"
     if [[ -f ${out} ]]; then
@@ -71,7 +73,8 @@ test: &test
     - run:
         name: Bring up the cluster
         command: |
-          build/portforward.sh 8080
+          export DIND_PORT_FORWARDER_WAIT=1
+          export DIND_PORT_FORWARDER=build/portforward.sh
           ./dind-cluster.sh up
     - run:
         name: Display cluster status (1)
@@ -83,7 +86,8 @@ test: &test
     - run:
         name: Bring up the cluster (again)
         command: |
-          build/portforward.sh 8080
+          export DIND_PORT_FORWARDER_WAIT=1
+          export DIND_PORT_FORWARDER=build/portforward.sh
           ./dind-cluster.sh up
     - run:
         name: Display cluster status (2)
@@ -130,8 +134,9 @@ test_src: &test_src
         name: Bring up the cluster
         no_output_timeout: 35m
         command: |
-          build/portforward.sh -wait 8080
-          build/portforward.sh -wait 8730
+          export DIND_PORT_FORWARDER_WAIT=1
+          export DIND_PORT_FORWARDER=build/portforward.sh
+          build/portforward.sh 8730
           cd kubernetes
           ../dind-cluster.sh up
     - run:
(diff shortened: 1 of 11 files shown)


hoegaarden:
@ivan4th
From our perspective this PR is ready. Please let us know what you think about
- having `dind-cluster.sh` start the portforwarder / have an option to call out to a user-defined executable as soon as the to-be-forwarded port is known
- the subnet discovery

Also, please let us know if you'd like us to squash the commits or such.
Thanks!

jellonek:
So this PR also incorporates a change in node naming with `DIND_LABEL` env usage, right?
If so - can you also add a note about that to PR title/description?

hoegaarden:
@jellonek The changes in node naming have already been introduced in #161 and we have documented that in the README. Let us know if the README isn't clear enough on that topic.

jellonek:
I was referring to change introduced in https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/161/files#diff-04c6e90faac2675aa89e2176d2eec7d8L66 and removed in this PR in https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/166/files#diff-04c6e90faac2675aa89e2176d2eec7d8L67.
If behavior was not changed from last PR, why this part of README was changed? Maybe it's only not needed?

hoegaarden:
@jellonek -- The behaviour regarding naming did not change since #161. In #161 the README was a bit misleading, as it suggests that every docker container's name is suffixed with the hash, which is not the case. When documenting the changes regarding subnets and ports for this PR, we decided to adapt the README and document the default case (no `DIND_LABEL` thus no suffix) and not the oder case with `DIND_LABEL` set.
Let us know if you'd prefer to split the doc update into a separate PR.

jellonek:
I'm perfectly fine with that, as it's already well described there :)

jellonek:
But for future it would be better to keep such changes in a separate PR.

hoegaarden:
@ivan4th: Do you maybe have any input on that PR? Is there anything left you want us to do for a *lgtm*?

ivan4th:
Thanks for the PR. The change is ok with me so I'm merging. One thing though... I'm not asking to do this before merging but maybe as a followup. We've got some inconsistent var/func naming in the code now. I mean `maxIP` instead of `max_ip` as a var name or `dind::clusterSuffix` instead of `dind::cluster-suffix` as a func name. Could you make another PR to fix these in the relevant parts of the code for the sake of consistency?

pmichali:
Late to the party, but I just saw this issue and corresponding PR. Several comments...

Since the behavior is changing, we should document in README.md the changes, or at least how it works now (have to specify API port and DinD subnet, or they will be randomized). I used it as-is, and it selected a subnet and API port, but the host is unable to access the API server using that port.

This change should address IPv6 also, not just IPv4 (otherwise, I guess multi-cluster will be broken for IPv6 mode). Would be nice if IPv6 wasn't a second class citizen.

Has this been tested in IPv6 mode, even with single cluster?

hoegaarden:
@pmichali 

> Since the behavior is changing, we should document in README.md the changes, or at least how it works now (have to specify API port and DinD subnet, or they will be randomized).

We tried to document the changes in the [PR](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/166#issue-201251452) and in parts of [the](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blame/32692857996406d3666726d53fd083884a791f7e/README.md#L226-L229) [README](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blame/32692857996406d3666726d53fd083884a791f7e/README.md#L164). Obviously we didn't do a great job. We will definitely try to do a better job!

> I used it as-is, and it selected a subnet and API port, but the host is unable to access the API server using that port.

I do understand the problem now (especially because your answer in the other [comment](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/186#discussion_r207225274). We seem to have different preferences & expectations:
- I like to use a specific `--context` when doing stuff
- I prefer tooling that does not mess with my defaults -- in that case the default context

So in this case it was clear to me, that it's the users responsibility to tell `kubectl`, or any other tool for that matter, which cluster I want to connect.

Having said that:
- I understand now, that this was a breaking UX change
- Add least would have needed (more) explicit docs 

In summary:
- We should merge #186 
- I / we should try to document that the current (after merge) UX
- In addition, to make UX (not only, but mostly) for multi cluster a bit better, I will think about a good way to switch / set / ... the context

> This change should address IPv6 also, not just IPv4 (otherwise, I guess multi-cluster will be broken for IPv6 mode). 

Yes. It should!
I don't have enough IPv6 context right now. But I am currently testing `kubeadm-dind-cluster` with IPv6 ...
It should not be totally broken with the changes #166 introduced -- BUT: there is a different UX (you need to explicitly set at least the `SUBNET` for multiple clusters. But we can def. try to fix that!

> Would be nice if IPv6 wasn't a second class citizen.

Well, then take away our IPv4 already! :smirk: 

> Has this been tested in IPv6 mode, even with single cluster?

As far as I can see there are no IPv6 tests, neither on Travis nor on CircleCI ...


Thanks in any case for your input!

pmichali:
@hoegaarden This commit has caused a regression with IPv6.

pmichali:
Replies to the above comments...

Unless I missed it, there is a behavior change, where the code now tries to assign a port and subnet, instead of providing a default and allowing the user to override. That needs to be documented in the README, IMHO.

It seems that you're looking at it from a multi-cluster perspective, where one is choosing which cluster to work with. In single cluster mode, the user shouldn't have to make those choices, it should just work without making selections.

I totally disagree with the summary suggestion (sorry).  Master is currently broken for IPv6, as of PR #166 (I filed PR #189 concerning that). We should not continue to apply commits to master. IMHO, we should revert #166, and resubmit it once it addresses IPv4 and IPv6, single and multi-cluster, and the change in behavior is documented and agreed upon by the community.

It is not a UX difference - PR #166 breaks IPv6 mode in single cluster and multi-cluster mode.

This is a major issue (no CI for IPv6) and an issue has been raised to track. For now, we need to manually check IPv6 mode in commits.

hoegaarden:
If you prefer rolling back over rolling forward (e.g. by merging #190 once it is confirmed to un-break IPv6 -- currently working on that), please do so! I def. want to unblock you asap!

pigmej:
> We should not continue to apply commits to master. IMHO, we should revert #166, and resubmit it once it addresses IPv4 and IPv6, single and multi-cluster, and the change in behavior is documented and agreed upon by the community

Can we list all PRs that have to land to unblock at least ipv6? Should we better roll back (revert) for a while or fixes are coming soon?

pmichali:
@pigmej My current plan, is to try to fix the current breakage, with the master branch, in its current state. I submitted PR #192 (most recent change yesterday - I've got to check the status). Once upstreamed, I plan on rebasing my fix (PR #186) to the first breakage for IPv6, At that point, IPv6 should be working.  Feel free to review those two PRs to quicken the process.

Issue #168, opened 2018-07-20
Create CI tests for new kube-router plugin


pmichali:
With the upstreaming of PR #159, kubeadm-dind-cluster now supports kube-router as a networking choice (for IPv4 mode).

Should revise the CI tests to run with CNI_PLUGIN=kube-router selected, to prevent regressions and verify operation.

PR #172, opened 2018-07-23
Add CI for kube-router networking option

pmichali:
Updates CircleCI to test with CNI_PLUGIN set to kube-router. Also
updated the README.md for kube-router and IPv6 in general.

Not sure if this should only be for K8s v1.11 or if it should be on
v1.10 as well (I added that for now).

Fixes Issue: #168


pmichali:
@ivan4th Here is the CI test for the recently added kube-router pod networking capability.

Issue #181, opened 2018-07-27
IPv6: kubectl broken, when no ~/.kube/config with use-context


pmichali:
Found an issue that, when I clear ~/.kube/config and then bring up the cluster in IPv6 mode, the cluster comes up fine, but unable to access via kubectl.

This is because, the context is set, but is not "in use". As a result, kubectl try to access API at localhost:8080, which doesn't work for IPv6 (It should be master node IP or ::1). If there is a ~/.kube/config with use-context: "dind" line, then it will work. Likewise, after the cluster is up, if the dind context is "used", then it will work.

Ivan indicated that the use-context line was removed from the script for multi-cluster mode. This breaks new runs with IPv6.

pmichali:
Fix has merged, so this is all set now.

PR #186, opened 2018-08-01
Missing use-context for non-multicluster

pmichali:
When not in multi-cluster mode, the set-context command must be
accompanied by the use-context command. Otherwise the host cannot
contact the cluster via API.

This change also renames the variable used to indicate whether the
docker daemon is running locally. Conditional tests can then
combine testing for GCE and remote docker daemon, as the both have
DOCKER_HOST set.

There are related/miscellaneous fixed as well...

For NAT64 we set a static route to for the IPv4 pool of the NAT64 server.
The remote docker case was not doing this route, so added in. Also, since
DOCKER_HOST is set for GCE or remote docker cases, the docker run command
can be used in all cases.

The DNS64 container used volume mounting, and created a config file on the
host in the /tmp area. This has several problems. First, a securty issue
with the temp file. Second, the file is copied over to a GCE container, but
not for a remote docker setup. To resolve this, the config is stored into
an environment variable and passed into the container at startup.

When running E2E tests, the API port is fixed to 8080. However, when running
under GCE, the API port is 8899 (and mapped to 8080 internally). This was
changed to use APISERVER_PORT for the test invocations.

Fixes issue: #181

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -43,11 +43,10 @@ if ! docker info|grep -s '^Operating System: .*Docker for Windows' > /dev/null 2
     fi
 fi
 
-# In case of linux as the OS and docker running locally we will be using
-# localhost to access the apiserver nor do we need to add routes
-using_linuxdocker=
+# Determine when using Linux and docker daemon running locally
+using_local_linuxdocker=
 if [[ $(uname) == Linux && -z ${DOCKER_HOST:-} ]]; then
-    using_linuxdocker=1
+    using_local_linuxdocker=1
 fi
 
 #%CONFIG%
@@ -673,9 +672,6 @@ function dind::ensure-volume {
 function dind::ensure-dns {
     if [[ ${IP_MODE} = "ipv6" ]]; then
         if ! docker inspect bind9 >&/dev/null; then
-	    local bind9_path=/tmp/bind9
-	    rm -rf ${bind9_path}
-	    mkdir -p ${bind9_path}/conf ${bind9_path}/cache
 	    local force_dns64_for=""
 	    if [[ ! ${DIND_ALLOW_AAAA_USE} ]]; then
 		# Normally, if have an AAAA record, it is used. This clause tells
@@ -685,7 +681,7 @@ function dind::ensure-dns {
 		# records meaning we ALWAYS use A records and do NAT64.
 	        force_dns64_for="exclude { any; };"
 	    fi
-	    cat >${bind9_path}/conf/named.conf <<BIND9_EOF
+	    read -r -d '' bind9_conf <<BIND9_EOF
 options {
     directory "/var/bind";
     allow-query { any; };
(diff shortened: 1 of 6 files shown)


dind-cluster.sh
```diff
@@ -16,6 +16,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 set -o errtrace
+set -x   # PCM temp
```
jellonek: Is that there leaved intentionally? Looks like debugging remains.

dind-cluster.sh
```diff
@@ -732,6 +716,10 @@ function dind::configure-kubectl {
     --server="http://${host}:${APISERVER_PORT}" \
     --insecure-skip-tls-verify=true
   "${kubectl}" config set-context "$context_name" --cluster="$cluster_name"
+  if [[ ${DIND_LABEL} = ${DEFAULT_DIND_LABEL} ]]; then
+      # Single cluster mode
+      "${kubectl}" config use-context "$context_name"
```
hoegaarden: This was removed, as also stated in the issue #181, for the multi cluster work. Because there are now potentially multiple `./dind-cluster.sh up` instances running in parallel which race for setting the default context -- so we decided to not do that and call `kubectl` with the context explicitly.

I am not sure why we need to set this in the single cluster mode, but I guess there is a good reason for that.

In case that would help: I was thinking about introducing a `context-name` sub-command to `dind-cluster.sh` (similar to the [`apiserver-port` sub-command](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/166/files#diff-6f53bfa951dee02d96a2466da8374ee8R1844)) which would allow external tools to query and use that.

dind-cluster.sh
pmichali: Yeah, forgot to pull. Will remove.

dind-cluster.sh
pmichali: In single cluster mode, we see that the context is not "in use". It is there, but not being used. As a result, from the host, one cannot access the API (kubectl commands fail). One could manually invoke the "use-context" command in that situation, but it then reduces the usefulness, as everything is not automated.

This issue appears in IPv6 mode, where the API server would be, for example, [::1], but when kubectl tried to access the API, it was using localhost, which maps to 127.0.0.1 on the host. Same with using the master IP, e.g. [fd00:77::2] in the set-context, the kubectl command was still trying to use localhost (because the context wasn't being used).

Ivan, suggested adding the use-context back in, conditionally,  so that it is only done for single cluster mode (line 719). This should allow multi-cluster to work, without the use-context, but continue to work for single cluster as well.

dind-cluster.sh
```diff
@@ -1353,15 +1341,15 @@ function dind::do-run-e2e {
...
+  if [[ ${using_linuxkit} || -z ${using_local_linuxdocker} ]]; then
     if [[ "${IP_MODE}" = "ipv4" ]]; then
       host="localhost"
     else
       host="[::1]"
     fi
   fi
   dind::need-source
-  local test_args="--host=http://${host}:${INTERNAL_APISERVER_PORT}"
+  local test_args="--host=http://${host}:${APISERVER_PORT}"
```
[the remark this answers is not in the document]
pmichali: Yeah, I hit these problems when trying to test E2E on GCE.

dind-cluster.sh
hoegaarden: Users need to set `DIND_LABEL` in any case if they want to create an additional cluster. If they don't, the script will just run `up` again on the already existing cluster.

I am totally OK with the change proposed in this PR. I am still struggling to understand the issue with not setting the default context, but that's probably just me.

In any case: Should we see to have an option to explicitly set the context to a specific cluster, something like:
```sh
./dind-cluster.sh up
./dind-cluster.sh set-context
kubectl ... # goes to the initial cluster
DIND_LABEL='another_cluster' ./dind-cluster.sh up
DIND_LABEL='another_cluster' ./dind-cluster.sh set-context
kubectl ... # goes to the second cluster
./dind-cluster.sh set-context
kubectl ... # goes to the initial cluster again
DIND_LABEL='another_cluster' ./dind-cluster.sh set-context
kubectl ... # goes to the second cluster again
```
This way, a user just needs to know about `DIND_LABEL` but not about any context name of any cluster.

pmichali: @hoegaarden For the failure, from a fresh system (no ~/.kube/ area), set IP_MODE=ipv6 and spin up a cluster. You will find that, from the host, you cannot access the API. Kubectl will indicate that it is trying to use localhost:8080, instead of [::1]:8080 or [fd00:10::2]:8080. Granted, you can force it to work, by changing /etc/hosts to map localhost to ::1, but that IMHO, is not a proper way to handle this issue.

Regarding multi-cluster, it probably would make sense (under a separate commit) to add documentation on how to use multicluster in the README.md file.

pmichali:
I'm rebasing to pick up #166, which has merged. Will rerun E2E tests, before pushing up a fix.

dind-cluster.sh
hoegaarden: I think most of the stuff should be addressed in https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/166#issuecomment-410007959 -- let me know if I have missed something!

pmichali:
This should NOT be merged, as the master branch is broken for IPv6 currently. We should not place this on top of a broken branch.

jellonek:
weave test looks like flaky - i've restarted the job.

pmichali:
@jellonek Seems like tide is acting flakey... currently, it says "Pending -- Not mergabe. Job ci/circleci: build has not succeeded.", but all tests have passed. On the last two commits, @ivan4th had to manually merge.

pmichali:
@jellonek Ah, looks like it just went in. Thanks!

Issue #182, opened 2018-07-27
IPv6: k-d-c CI needed for IPv6 mode...


pmichali:
Given the several regressions we've seen in IPv6 mode lately, it would be great to have CI tests to check regressions in IPv6 mode operation. Testing should cover things like:

- bridge CNI plugin mode.
- kube-proxy with IPVS.
- (future) PTP CNI plugin mode.
- Accessing remote systems using NAT64, for both systems that support IPv6 and IPv4
- When allowing AAAA use (direct access to external IPv6 hosts w/o NAT64)
- Pop to pod access, across nodes

pmichali:
Additional IPv6 tests...
- up/down/up cycle and check pod-to-pod pings and pod to external access.
- Check that tayga/bind9 containers are still running, after up/down.
- Check that tayga/bind9 containers not running, after "clean".
- Negative: multi-cluster and IPv6, IPv6 with kube-router
- Clsuter up with CoreDNS and check DNS for pods and external IPs (V4 and V6, when have allow AAAA mode).

Could use nslookup on pod for checking AAAA vs synthesized IPv6 address with embedded IPv4 (using DIND_ALLOW_AAAA_USE flag, and check that IP doesn't have DNS64 prefix).

PR #190, opened 2018-08-02
Introduce tests for IPv6, update documentation about testing/contributing

hoegaarden:
Introduces some initial tests for IPv6 to run locally and on CircleCI

Implements part of: #182


hoegaarden:
I believe the failing test is a flake, however I am not allowed to rerun it.

pmichali:
There are additional issues, not being addressed... ensure-network() calls get-subnet(), which does not correctly parse the docker network, when in IPv6 mode. There will be two subnets from the docker inspect (one IPv4, one IPv6, in any order). 

ensure-network() does not set the GW for IPv6 mode correctly.

Nit: why do an inspection of the docker network, every time an I address is needed on the DinD subnet? Why not cache the value and use it throughout the script, like what was originally there (only programmable)?  IOW, instead of using a fixed default and allowing config override, like the original, calculate an available IPv4 subnet and store it in DIND_NET and DIND_SUBNET_SIZE, again allowing override via config, if desired. Use that subnet prefix everywhere, when creating IPs, like the original code.  I don't see the need to dynamically determine the subnet.

pmichali:
I'm assuming you are disabling IPv6 for multi-cluster, because each cluster needs a different DinD subnet?  I think it is OK to put a quick test at the start of the script to exit, if v6 and multi-cluster for now.

For IPv6 and multicluster, if you can introduce an ID number for each cluster (or generate a random one), it could be incorporated into the IP address easily... For example, instead of the default fd00:10::/64, it could be fd00:10:X::/64, where x=1..ffff.

I'm playing with modifying the current code to work with IPv6 for single cluster. If I get something, I can post diffs.

hoegaarden:
- We added one test to run on TravisCI which tests internal IPv6 connectivity & external via NAT64 (`test-case-ipv6-aaaa-disallowed`).
- For direct external IPv6 connectivity (`test-case-ipv6-aaaa-allowed`) we'd need to have a hosted CI with IPv6 enabled. Neither TravisCI nor CIrcleCI seem to offer that -- in fact, we couldn't find any hosted CI which supports IPv6.
- We added a section in the README describing how to run certain tests locally and which CI systems are in place.

We will continue to look into having a proper IPv6 CI, but this is at least a start.

@pmichali 
- Do you think the tests are OK for now or are we missing something important that could be tested even in a non -IPv6 environment?
- As soon as #192 is merged, we will rebase onto that.

Addresses parts of: #182

README.md
```diff
@@ -306,6 +308,39 @@ currently have some issues. You may still try running them though:
...
+
+There isn't currently a CI setup with IPv6 support available. Please run at
+least the following locally, even if your contribution was not directly related
+to IPv6:
+```shell
+# should succeed on any environment
+$ TEST_CASE=ipv6-aaaa-disallowed ./test.sh
+# needs IPv6 enabled environment to succeed
+$ TEST_CASE=ipv6-aaaa-allowed ./test.sh
+```
```
pmichali: FYI,  One should be able to enable AAAA mode, and even though they will be unable to ping an IPv6 capable external system given their network setup, and then do nslookup on pod and see that some external sites (e.g. google.com) should come up with an IPv6 address that does not have the DNS64 prefix, and others (e.g. github.com) should have the prefix.

May be a way to test both w/o running multiple tests.

dind-cluster.sh
```diff
@@ -265,6 +265,13 @@ fi
 DEFAULT_DIND_LABEL='mirantis.kubeadm_dind_cluster_runtime'
 : "${DIND_LABEL:=${DEFAULT_DIND_LABEL}}"
 
+# TODO -- make multi cluster && ipv6 work
```
pmichali: This is in PR 192 (which I hope upstreams VERY soon).

test.sh
```diff
@@ -308,6 +308,120 @@ function fail() {
...
+  echo '{ "apiVersion": "v1", "spec": { "nodeSelector": { "kubernetes.io/hostname": "'"$node"'" } } }'
+}
+
+function get-hostname-override() {
+  local name="${1}"
+  echo '{ "apiVersion": "v1", "spec": { "hostname": "'"$name"'" } }'
+}
+
+function setup-ipv6-test() {
+  # Currently only testing with a specific version of
```
pmichali: Ideally we want to test with 1.9+ versions

test.sh
```diff
@@ -308,6 +308,120 @@ function fail() {
...
+  $k run pod-on-node1 --image=busybox --restart=Never \
+    --overrides="$( get-node-selector-override kube-node-1 )" \
+    --command -- /bin/sh -c 'hostname ; /bin/sleep 3600'
+
+  local ipv6_enabled_host="ds.test-ipv6.noroutetohost.net"
+  local ipv4_only_host="ipv4.test-ipv6.noroutetohost.net"
+
+  pod_on_node1_ip="$( get-ip-for-pod 'pod-on-node1' )"
+
+  $k exec -ti pod-on-node1 -- ping6 -c1 $ipv6_enabled_host || {
```
pmichali: Mentioned above, it may be easier to do nslookup on the pod for google.com and github.com and check that the former does NOT start with the DNS64 prefix and the latter does start with it.

test.sh
```diff
@@ -308,6 +308,120 @@ function fail() {
...
+
+  ping-tests
+
+  IP_MODE='ipv6' \
+    bash -x ./dind-cluster.sh clean
+}
+
+# with DIND_ALLOW_AAAA_USE set
+function test-case-ipv6-aaaa-allowed() {
+  export DIND_ALLOW_AAAA_USE=some-value
```
pmichali: use true, instead of some-value.

test.sh
```diff
@@ -308,6 +308,120 @@ function fail() {
...
+    bash -x ./dind-cluster.sh up
+
+  ping-tests
+
+  # additional assertions
+  local k="kubectl --context dind"
+  local ipv6_host='ds.test-ipv6.noroutetohost.net'
+  local ipv6_host_ns='8.8.8.8'
+
+  expected_ipv6_addr="$( dig +short "@${ipv6_host_ns}" "${ipv6_host}" aaaa )"
```
pmichali: Although it may be nice to actually ping to an IPV6 destination, I'm wondering if that will be problematic, as some sites, like google.com have multiple IPv6 addresses.  A compromise may be to just do a ping (by name) to an IPv4 site, and ensure that it works (and that NAT64 is working), and then use the nslookup mentioned before to verify that DNS64 is working for both AAAA and non-AAAA destinations.

test.sh
```diff
@@ -308,6 +308,120 @@ function fail() {
...
+function test-case-ipv6-aaaa-disallowed() {
+  setup-ipv6-test
+
+  IP_MODE='ipv6' \
+    bash -x ./dind-cluster.sh up
+
+  ping-tests
+
+  IP_MODE='ipv6' \
+    bash -x ./dind-cluster.sh clean
```
pmichali: Would be good to do a down, verify that tayga and bind9 containers are still running, do an up, and make sure that ping test works again, and then do the clean and make sure tayga and bind9 containers are NOT running.

pmichali:
See PR #192, which has merged, as some of the code changes in dind-cluster.sh are handled.  Also, please see Issue #182 for ideas on IPv6 testing for CI.

test.sh
hoegaarden: Sure. I want to get the test case(s) in place first. Having them run for multiple versions should then be a rel. easy addition.

hoegaarden:
@pmichali PTAL

We are aware, that we don't test every aspect of your list yet, but we'd like to get that out sooner then later and maybe add the additional  cases from #182 and potentially others one by one.

@mariantalla

hoegaarden:
@pmichali -- question regarding https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/190#discussion_r209030154 :

> Ideally we want to test with 1.9+ versions

We saw that `kube-dashboard` and `kube-dns` pods were failing for `v1.8` after the snapshotting `k-d-c` does. We didn't investigate too much yet, partially also because you seem to be OK with not supporting 1.8.

Could you please add some context about IPv6 and 1.8: Is it not (well) supported? Is it that we just don't care about more the `n-2`?

README.md
```diff
@@ -322,6 +322,51 @@ currently have some issues. You may still try running them though:
...
 ```
 
+## Contributing to & Testing `kubeadm-dind-cluster`
+
+There are currently two CI systems in place which automatically test PRs to
+kubeadm-dind-cluster:
+- [TravisCI](https://travis-ci.org/kubernetes-sigs/kubeadm-dind-cluster)
+- [CircleCI](https://circleci.com/gh/kubernetes-sigs/kubeadm-dind-cluster)
+
+To run tests locally, find the test function you wish to run (it is called
```
pmichali: Can you check with @ivan4th? I thought he mentioned that test.sh was being deprecated and that we should be doing CircleCI tests only.

README.md
```diff
@@ -322,6 +322,51 @@ currently have some issues. You may still try running them though:
...
+
+#### `TEST_CASE=ipv6 DIND_ALLOW_AAAA_USE=true ./test.sh`
+
+Those tests use the public AAAA records when available. This means that for
+hosts which have a AAAA record that IP is used, only in case a host does not
+have a AAAA record, the DNS64 kicks in. Furthermore that means, that traffic to
+those host does not get routed through NAT64. In that case the host running the
+outer docker daemon needs to have external IPv6 available.
+
+As our CI currently does not allow for external IPv6 the external ping tests
```
pmichali: We could, for this case, just check that name lookup works correctly. Namely, that for external sites that support IPv6, the address should NOT have the DNS64 prefix and external sites that only support IPv4 will always have the DNS64 prefix (and use an embedded IPv4 address). I wouldn't advocate checking the IPv4 address, but only whether or not the prefix exists, in these too cases.

test.sh
```diff
@@ -308,6 +308,237 @@ function fail() {
...
+  $kubectl run pod-on-node1 --image=${k8s_img} --restart=Never \
+    --overrides="$( get-node-selector-override kube-node-1 )" \
+    --command -- /bin/sh -c 'hostname ; /bin/sleep 3600' \
+    >/dev/null 2>&1
+
+  echo "$( get-ip-for-pod 'pod-on-node1' )"
+}
+
+function ping-tests-external() {
+  if [ -n "${DIND_ALLOW_AAAA_USE:-}" ]
```
pmichali: mentioned above, could just check name lookup for IPv6 capable external hosts, but actually do a ping to IPv4 hosts, rather than exist the test, if set.

test.sh
```diff
@@ -308,6 +308,237 @@ function fail() {
...
+function get-addr-on-pod() {
+  local pod="$1"
+
+  local name="$2"
+  local type="${3:-A}"
+  local ns="${4:-}"
+
+  test -n "$ns" && ns="@${ns}"
+
+  $kubectl exec -ti "$pod" -- dig +short "${name}" "${type}" ${ns} | sed 's/\r//g'
```
pmichali: For pods, should just be able to use "ip a" or ifconfig for eth0. I wouldn't think there would be an A record for the pods. should only have IPv6 address.

test.sh
```diff
@@ -308,6 +308,237 @@ function fail() {
...
+  setup-ipv6-test
+
+  IP_MODE='ipv6' \
+    bash -x ./dind-cluster.sh up
+
+  persistent_pod_ip="$( create-persistent-pod )"
+
+  for target in "${ipv6_enabled_host}" "${ipv4_only_host}"
+  do
+    aaaa_from_node="$( get-addr-on-node 'kube-node-1' "${target}" AAAA )" || {
```
pmichali: Not sure of the usefulness of getting the IPs for the nodes. Goal is to ensure pod to pod access.

test.sh
```diff
@@ -308,6 +308,237 @@ function fail() {
...
+      fail "Expected to successfully resolve ${target}'s IPv6 address on kube-node-1"
+    }
+    a_from_node="$( get-addr-on-node 'kube-node-1' "${target}" A )" || {
+      fail "Expected to successfully resolve ${target}'s IPv4 address on kube-node-1"
+    }
+
+    aaaa_from_pod="$( get-addr-on-pod 'pod-on-node1' "${target}" AAAA )" || {
+      fail "Expected to successfully resolve ${target}'s IPv6 address on a pod"
+    }
+    a_from_pod="$( get-addr-on-pod 'pod-on-node1' "${target}" A )" || {
```
pmichali: I would think that the pods should not have an IPv4 address, and only should have an IPv6 address.

pmichali:
@hoegaarden IPv6 support was added in K8s 1.9 (as alpha feature), so it is not supported at all in 1.8 and older.

test.sh
hoegaarden: I see the confusion: `get-addr-on-pod` functions are not about getting the address of a pod, it is about resolving a name from within the pod. We can rename the function(s) to make their purpose clearer.

test.sh
hoegaarden: pod to pod (ping) is tested [here](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/190/files#diff-1b531c5e405e0bc95890d52cf2f2f5b5R436). This is testing the DNS resolution of a given name on a pod.

test.sh
hoegaarden: Same as [below](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/190#discussion_r210202524): This is a bad name, it actually tests DNS resolution from within the pod.

README.md
hoegaarden: It seems that this is what the tests are doing at the moment, if I've got this right. Are there points that you'd rather see implemented differently?

test.sh
hoegaarden: Is this covered by [this](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/870ba8c4a5c5bf0289a97ef78710f38f2295497e/test.sh#L501-L526) ?

test.sh
pmichali: Maybe break the logic out into a function that indicates it is testing name lookup and call it, like you are doing in L529-530 with the ping tests, so that test-case-ipv6 is at a higher level, calling 3 tests, internal ping, external ping, and name lookup. Later more can be added.

test.sh
```diff
@@ -308,6 +308,237 @@ function fail() {
...
+    bash -x ./dind-cluster.sh up
+
+  persistent_pod_ip="$( create-persistent-pod )"
+
+  for target in "${ipv6_enabled_host}" "${ipv4_only_host}"
+  do
+    aaaa_from_node="$( get-addr-on-node 'kube-node-1' "${target}" AAAA )" || {
+      fail "Expected to successfully resolve ${target}'s IPv6 address on kube-node-1"
+    }
+    a_from_node="$( get-addr-on-node 'kube-node-1' "${target}" A )" || {
```
pmichali: Since this is IPv6 only mode, we won't ever be accessing hosts by IPv4 addresses, so tests like L461-463, L468-L470, L475-477,... are not really needed.

README.md
pmichali: Did you get to touch base with Ivan?

README.md
```diff
@@ -322,6 +322,51 @@ currently have some issues. You may still try running them though:
...
+$ TEST_CASE=<test-name> ./test.sh
+```
+
+### IPv6 tests
+
+All of the IPv6 related tests currently run on
+[TravisCI](https://travis-ci.org/kubernetes-sigs/kubeadm-dind-cluster). There
+are two slightly different kind of tests which run for all major version starting from `v1.9`:
+
+#### `TEST_CASE=ipv6 ./test.sh`
```
pmichali: Consider calling the test case "ipv6-only", as in the future, we'll have dual-stack support support in K8s and maybe it'll reduce confusion?

hoegaarden:
@pmichali -- *PTAL* and let us know if that test setup looks good to you

- We refactored the test a bit based on your comments
- We dropped Ivan a message about Travis vs. CircleCI vs. `./test.sh` -- we will let you know once we know where this is going

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+function setup-ipv6-only-test() {
+  export IP_MODE='ipv6'
+
+  # set TEST_K8S_VER to something specific, by default testing v1.11 for now
+  local v="${TEST_K8S_VER:-v1.11}"
+  export LOCAL_KUBECTL_VERSION="$v"
+  export DIND_IMAGE="mirantis/kubeadm-dind-cluster:${v}"
+  docker pull "${DIND_IMAGE}"
+
+  # Globally exported CIDR not applicable for IPv6
```
pmichali: Don't need to clear this, as dind-cluster.sh will set to default, if not specified by user.

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+  export POD_NETWORK_CIDR=''
+  # Add the directory we will download kubectl into to the PATH
+  export PATH="${PATH}:${KUBECTL_DIR}"
+
+  # Explicitiely set the DNS64 prefix
```
pmichali: This is defaulted too, in dind-cluster.sh

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+  export DNS64_PREFIX='fd00:10:64:ff9b::'
+
+  # some shared config for the ipv6 tests
+  export ipv6_enabled_host="ds.test-ipv6.noroutetohost.net"
+  export ipv4_only_host="ipv4.test-ipv6.noroutetohost.net"
+  export ipv6_enabled_host_ns='8.8.8.8'
+
+  # kubectl setup
+  export kubectl="kubectl --context dind"
```
pmichali: I'm pretty sure this is no longer needed either, as the context has been fixed for single cluster use.

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+
+  # the image to use for containers in k8s
+  export k8s_img='tutum/dnsutils'
+}
+
+function lookup-address() {
```
pmichali: would dns-lookup-ip-by-type be clearer?

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+  local name="$2"
+  local type="${3:-A}"
+  local ns="${4:-}"
+
+  test -n "$ns" && ns="@${ns}"
+
+  $kubectl exec -ti "$pod" -- dig +short "${name}" "${type}" ${ns} | sed 's/\r//g'
+}
+
+function lookup-address-on-node() {
```
pmichali: lookup-ip-address-on-node or lookup-node-ip?  Just trying to reduce the ambiguity of "address" on these function names.

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+    --overrides="$( get-node-selector-override kube-node-1 )" \
+    --command -- /bin/sh -c 'hostname ; /bin/sleep 3600' \
+    >/dev/null 2>&1
+
+  echo "$( get-ip-for-pod 'pod-on-node1' )"
+}
+
+function ping-tests-external() {
+  $kubectl exec -ti pod-on-node1 -- ping6 -n -c1 $ipv6_enabled_host || {
+    fail "Expected ping6 to $ipv6_enabled_host to succeed"
```
pmichali: Curious... if the test fails here and other checks, will the pod(s) be cleaned up?

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+  local name="$1"
+  local type="${2:-A}"
+  local ns="${3:-}"
+
+  test -n "$ns" && ns="@${ns}"
+
+  dig +short ${name} ${type} ${ns} | sed 's/\r//g'
+}
+
+function lookup-address-on-pod() {
```
pmichali: would dns-lookup-ip-on-pod-by-type be clearer?

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  # We verified, that AAAA is the same from Pod & Node, so we can just return
+  # one of those
+  local aaaa_from_k8s="${aaaa_from_pod}"
+
+  echo "${aaaa_from_host},${aaaa_from_k8s}"
+}
+
+function verify-lookup-with-dns64-only() {
```
pmichali: verify-lookup-always-using-dns64-prefix?

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+    fail "Expected ${target} to resolve to something starting with the DNS64 prefix on pods/nodes, but got ${aaaa_from_k8s}"
+  fi
+
+  if starts-with "$DNS64_PREFIX" "$aaaa_from_host"
+  then
+    fail "Expected ${target} not to resolve to something starting with the DNS64 prefix on the host"
+  fi
+}
+
+function verify-lookup-with-ext-dns-for-ipv6-host() {
```
pmichali: verify-lookup-using-ipv6-address-for-ipv6-hosts?

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+    fail "Expected ${target} not to resolve to something starting with the DNS64 prefix on the host"
+  fi
+
+  if [ "$( lookup-address "$target" 'AAAA' "$ipv6_enabled_host_ns" )" != "$aaaa_from_k8s" ]
+  then
+    fail "Expected ${target} to resolve to the same as on ${ipv6_enabled_host_ns}"
+  fi
+}
+
+function verify-lookup-with-ext-dns-for-ipv4-only-host() {
```
pmichali: verify-lookup-using-dns64-prefix-for-ipv4-hosts?

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+  if [ "$aaaa_from_host" != "" ]
+  then
+    fail "Expected ${target} not to resolve on the host"
+  fi
+}
+
+function test-case-ipv6-only() {
+  setup-ipv6-only-test
+
+  bash -x ./dind-cluster.sh up
```
pmichali: Using -x for debug purposes?

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+
+  ping-tests-internal "$persistent_pod_ip"
+
+  if [ -z "${DIND_ALLOW_AAAA_USE:-}" ]
+  then
+    ping-tests-external
+  else
+    {
+      echo ''
+      echo 'WARNING: not testing direct external IPv6 connectivity.'
```
pmichali: Maybe a NOTE vs WARNING?

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,244 @@
...
+      echo '  ... right now IPv6 in CI is only tested through NAT64 as'
+      echo '      we do not have access to an IPv6 enabled CI system.'
+      echo ''
+      echo '  ... "ping-tests-external" is skipped'
+      echo ''
+    } >&2
+  fi
+
+  bash -x ./dind-cluster.sh clean
```
pmichali: -x for debug? Pulling later?

test/test-ipv6-only.sh
hoegaarden: This is explicitly set, so in the tests we can check if a record is prefixed with `DNS64_PREFIX`. If we'd leave it to `./dind-cluster.sh` to default that, we couldn't really use that in the test assertions. I'll try yo add a better comment describing that.

test/test-ipv6-only.sh
hoegaarden: This is an artifact from having those tests in `./test.sh` ... I will remove that.

test/test-ipv6-only.sh
hoegaarden: `-x` because this was the default in `./test.sh` and I just went with it. I find it nice but I am happy to remove it to be more consistent with other CIrcleCI test or if people prefer to not use this.

test/test-ipv6-only.sh
hoegaarden: regarding `-x` see https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/190#discussion_r210905163

> Pulling later?

I am not sure what you mean by that? Could you please elaborate?

test/test-ipv6-only.sh
pmichali: Gotcha, was thinking the setting from dind-cluster.sh was visible.

test/test-ipv6-only.sh
pmichali: It displays output that is useful during debug, but not sure we need all that output for normal runs. You could use a variable DEBUG="-x", so it could be turned on/off, if there are problems with running dind-cluster.sh.

test/test-ipv6-only.sh
pmichali: The -x just shows line by line execution of the script. Useful for debug, but I don't know if you need the option in the final run, especially since it is just teardown.

test/test-ipv6-only.sh
hoegaarden: Nope, when any test fails, no cleanup will be done and the test is aborted right away. This is, at least from my point of view, deliberate: In CI everything will be torn down anyway, locally this gives the user a chance to debug.

hoegaarden:
@pmichali *PTAL*

README.md
```diff
@@ -322,6 +322,89 @@ currently have some issues. You may still try running them though:
...
+* internal `ping6`s (pod to pod on different nodes)
+* external `ping6`s (to IPv4-only and IPv6-enabled targets)
+
+#### `TEST_K8S_VER='1.x' DIND_ALLOW_AAAA_USE=true ./test/test-ipv6-only.sh`
+
+Those tests use the public AAAA records when available. Specifically for hosts
+which have a AAAA record, the IP address is used, traffic to those hosts does
+not get routed through NAT64. In that case the host running the outer docker
+daemon would need to have external IPv6 available to actually communicate with
+external IPv6 hosts. Therefore (becuase none of our CI systems can provide
```
pmichali: nit: becuase -> because

hoegaarden:
@pmichali Thanks for your help on that!

@ivan4th / @pigmej Could you maybe also check the PR and give the *lgtm* if you are happy with it?

test/test-ipv6-only.sh
```diff
@@ -0,0 +1,248 @@
...
+  kubectl run pod-on-node1 --image=${k8s_img} --restart=Never \
+    --overrides="$( get-node-selector-override kube-node-1 )" \
+    --command -- /bin/sh -c 'hostname ; /bin/sleep 3600' \
+    >/dev/null 2>&1
+
+  echo "$( get-ip-for-pod 'pod-on-node1' )"
+}
+
+function ping-tests-external() {
+  kubectl exec -ti pod-on-node1 -- ping6 -n -c1 $ipv6_enabled_host || {
```
pmichali: Just curious as to whether the ping test should be with a count of 5 vs 1 (in case there is some network glitch that causes one ping to fail)? I'm not sure what happens if several pass and a few fail, w.r.t return code.

hoegaarden: I'd say we keep it as is (one ping). We had it running quite often now (locally, on Travis & on CircleCI) and one ping was always fine. If we see that those tests are getting flaky we can see if we want to use multiple pings.

README.md
```diff
@@ -322,6 +322,89 @@ currently have some issues. You may still try running them though:
...
+case also locally, by just calling the script:
+
+```shell
+$ DIND_ALLOW_AAAA_USE='true' TEST_K8S_VER='v1.10' ./test/test-ipv6-only.sh
+```
+
+#### TravisCI, `./test.sh` -- DEPRECATED
+
+There are some tests in `./test.sh`. These were the tests which ran on TravisCI
+when TravisCI was the default CI system.
```
ivan4th: Small correction, sorry, perhaps I was not precise enough when I was trying to describe my idea of using TravisCI. TravisCI is still supported for people who want to use kdc there, but only a few cases are run there now instead of the full suite. The corresponding unused parts of test.sh can be removed. So test.sh / TravisCI is not completely deprecated, just parts of it.

.travis.yml
```diff
@@ -14,3 +14,6 @@ before_install:
 script:
   # building k8s may cause Travis to time out w/o output
   - if [ -n "${TWAIT:-}" ]; then travis_wait 35 ./test.sh; else ./test.sh; fi
+
+before_script:
+  - echo 0 | sudo tee /proc/sys/net/ipv6/conf/all/disable_ipv6
```
ivan4th: We don't test IPv6 on Travis, so why do we need this?

fixed/dind-cluster-stable.sh
```diff
@@ -1152,9 +1152,9 @@ function dind::setup_external_access_on_host {
   fi
   local main_if=`ip route | grep default | awk '{print $5}'`
   local bridge_if=`ip route | grep ${NAT64_V4_SUBNET_PREFIX}.0.0 | awk '{print $3}'`
-  docker run --entrypoint /sbin/ip6tables --net=host --rm --privileged ${DIND_IMAGE} -t nat -A POSTROUTING -o $main_if -j MASQUERADE
```
ivan4th: `dind-cluster-stable.sh` is supposed to be used with `mirantis/kubeadm-dind-cluster:stable` image and is only updated when we need to update that stable image. We're not updating it right now, so perhaps we should leave `fixed/dind-cluster-stable.sh` untouched.

README.md
hoegaarden: Got it -- will update the docs on that.

fixed/dind-cluster-stable.sh
hoegaarden: - that's a change done by `genfixed.sh`
- CI would fail if we didn't have that change
- This code was there before already as is, it has just been pushed into the function [`dind::ip6tables-on-hostnet`](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/b91d8899a3bee39683f427b855fb34866f006fee/fixed/dind-cluster-stable.sh#L1190)
- `DIND_IMAGE` is set specifically to `mirantis/kubeadm-dind-cluster:stable` [here](https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/b91d8899a3bee39683f427b855fb34866f006fee/fixed/dind-cluster-stable.sh#L52)

With all that, could you more context what's the issue with this change? Am I missing something?

pmichali: I thought this change was just pulling out duplication and placing into a common function. It should be a refactoring and not altering the script functionality, right?

ivan4th:
Thanks for the PR!

Issue #189, opened 2018-08-02
Regression(s) from PR166


pmichali:
PR 166 changes API server port to be dynamic, and also calculates the IPv4 subnet to use for the DinD network.

What I see for IPv6, is that, when the dind-network is created, it computes a V4 subnet (OK), uses the specified IPv6 network (OK), but tries to set the GW based on the V4 subnet and not the V6 subnet.

As a result, we see:
```docker: Error response from daemon: invalid IPv6 address: 0.0.0.2.```

With IPv4, I'm also having issues with the cluster coming up and accessing the API **from the host**, in bare-metal single cluster mode. From the host, I see kubectl is trying to access the API at the master node's IP and a dynamic port. From the master node, the API server is running on localhost and the dynamic port. Looking at netstat, the host is listening on 127.0.0.1 and the dynamic port.

If I set APISERVER_PORT to 8080, the cluster comes up, and I can access the host.

pmichali:
There are several functions that do not support IPv6: get-subnet, get-ip-from-rnage, ensure-net,...

PR #192, opened 2018-08-03
IPv6: fix breakage from PR 166

pmichali:
This change will disable multi-cluster use with IPv6. Instead of
trying to look up subnet everytime an address is generated, for
IPv4, a subnet is generated (as before), and stored in a variable
for later use. This averts the issue of trying to determine the
subnet, when using IPv6 (where there are multiple subnets).

Regardng generation of IPs, the IPv4 generation remains the same.
For IPv6, the IP is just added as the least significant 16 bits.

Note: Not sure if the existing IPv4 address generation takes into
account avoiding anycast/multicast IPs.

Also, GW is set to the proper subnet.

Note: A a separate commit, the IPv6 DinD subnet could be generated
(e.g. adding some form of cluster ID to subnet network part - as in
fd00:10:xxxx::/64) and then can be tested with multi-cluster.

Fixes: #189


jellonek:
You need to rebase this PR on top of current master to have `kube_router` tests unblocked.

pmichali:
Checking to see if kube-router tests pass, now that rebased on master.  Need to see why one multi-cluster test failed for 1.10, but worked with other versions. There were two src tests that failed in test setup.

If this works, will re-push with correction to commit message, which currently is:

```
    IPv6: Addressing breakage and related issues

    There are several issues found:

    1) Every time an address is needed on the DinD network, a lookup
       is done to determine the CIDR. This is inefficient, and it
       functionality is multi-purpose in that it creates the network,
       when it doesn't exist.
    2) If the DinD network exists, the parsing of the subnet fails,
       when IPv6 is involved (as there are multiple subnets).
    3) Because of parsing issues, the gateway address calculation is
       wrong in IPv6 mode (uses IPv4 address).
    4) Unrelated to PR 166, the NAT64 and DNS64 containers are
       deleted on "down" and restored on "up", dind::run will have
       already created these containers and duplicate containers
       occur.
    5) IPv6 hasn't been tested for multi-cluster, and needs to be
       checked (and likely modified to dynamically create pod subnet
       based on the cluster instance), before being enabled.
    6) The kube-router CI tests were failing, as well.

    For (1), we'll only determine the IP/prefix for the DinD network
    once at startup. For IPv6, this is unchanged, it is either a
    provided env variable, or default is used. For IPv4, it will
    use the IP/prefix from the existing network, will use a provided
    env variable value, or determine an available CIDR to use for the
    DinD network and assign that to a variable for later use. This
    separates network definition from creation.

    For (2), the parsing is modified, to account for the possiblity
    of an existing DinD network with an IPv6 subnet. This parsing
    will only be done for IPv4 currently. In the future, IPv6 could
    be parsed, and if no DinD network, an address can be calculated,
    when not provided.

    For (3), because we are using variables for the DinD network IP
    and mask, and not looking up the values (and failing in the
    parsing), the gateway assignment is trival.

    For (4), during testing it was seen that NAT64 and DNS64 have the
    same label as node containers, causing them to be deleted. The
    solution is to use a different label, and only delete them during
    a "clean" operation. The original goal was to reduce startup time,
    by not creating these containers all the time. One a related note,
    they likely can be shared by multiple clusters, so we don't want
    to delete them during down.

    For (5), a test was added at statup for IPv6 and mult-cluster, and
    the script exited.

    For (6), needed rebase with master to pick up fix to set context
    for kube-router use. Update README.md to indicate how to use
    kube-router.

    FUTURE: The network name is computed on every call that needs
    to reference the kubeadm-dind-net. This too, could be done one
    and stored in a variable.

    FUTURE: Could add a cluster ID to the network address use for
    the kubeadm-dind-net in IPv6 mode to allow multiple cluster support.

    FUTURE: When IPv6 tests are added to CI, beside testing startup
    and operation, the up/down/up cycle should be checked, along with
    persistence of NAT64/DNS64 containers.
```

dind-cluster.sh
```diff
@@ -173,6 +229,9 @@ if [[ ${IP_MODE} = "ipv6" ]]; then
 	exit 1
     fi
 else
+    dind::find-or-create-ipv4-dind-subnet
+    dind_ip_base="$(echo "${DIND_SUBNET}" | sed 's/0$//')"
```
hoegaarden: nit: I think this is not needed anymore -- is it?

pmichali: L233 isn't needed. I can remove. However, I'd posit that using dind_ip_base, is a much simpler mechanism for fabricating an IP address from a subnet prefix.  The only time we really need logic like that which is used in get-ip-from-range() for IPv4, is when we expect to have more than 255 IPs and need to deal with rollover.

I really doubt we'd try to create a cluster with over 10 nodes, let alone over 255. IMHO, I think we added a lot of complexity, for little value. The same goes for the hash used to identify different clusters (label). I would be surprised if we needed, in a practical sense,  a large number of clusters that warrant a hash.

Besides the added complexity, it makes it much hard to build subnets (both IPv4 and IPv6). I would think we could use a simple integer for the cluster ID and use it in labels, and in IPv4 and IPv6 addresses.

For example, we could just use "fd00:<cluster#>:20::/64" for an IPv6 DinD subnet, where <clsuter#> is 1..N.

hoegaarden: > L233 isn't needed.

I was talking about L233 -- no real need to remove it right now.

> [...] The only time we really need logic like that which is used in get-ip-from-range() for IPv4, is when we expect to have more than 255 IPs and need to deal with rollover. [...]

It is not only a problem for a huge number of nodes, but also if other `DIND_NETWORK_SIZE`es are in play and if we need to find out how big the already allocated docker networks are, and where we can place a new network. I think this is the only way to somehow reliably do that -- If someone has a better idea, I am duper happy to do something simpler.

pmichali: I just want to be clear, as you mentioned L233, and I agreed that L233 is not needed, but you mentioned again about L233. Aren't we in violent agreement that L233 is not needed? Or are you talking about L232?

Regarding using the subnet prefix and simply appending the interface number, my point was that we should follow YAGNI, and only add the complexity, when we have an actual use case needing it. Do we have a use case right now?

hoegaarden: Sorry for the confusion: L232 is def. needed and needs to be in the PR. L233 could be removed (but does not do harm if left in).

The whole network discovery / calculation was implemented, because when we implemented that, we had some networks configured on docker already and we ran into the problem with a simple approach you described. Therefore we decided to not make any assumptions and about e.g. networks already configured on users' machines. Another thing that was in the back of our heads: We were thinking about making k-d-c a backend/implementation of the [test cluster framework](https://github.com/pivotal-k8s/testing_frameworks/tree/introduce-test-cluster-framework/dind) which should allow "many" parallel running clusters. Also we were under the impression, that `DIND_SUBNET_SIZE` was configurable for a reason.

Having said that: I am happy to go with a simpler solution! E.g., as you described already, allocate `10.192.$N.0` / `fd00:$N:20::` and fix the size to `24` / `64`.

jellonek:
Will look on them on the morning. Looks like something there was an issue with upstream sources compilation, so I'm almost sure that failure was disconnected from changes in this PR.

jellonek:
Looks like that was temporary issue with building k8s from master, now all tests are green.

jellonek:
Ups, I see that you commented out these tests. Please uncomment them.

README.md
```diff
@@ -122,6 +122,15 @@ The first `dind/dind-cluster.sh up` invocation can be slow because it
 needs to build the base image and Kubernetes binaries. Subsequent
 invocations are much faster.
 
+## Kube-router
+Instead of using kube-proxy and static routes (with bridge CNI plugin),
+kube-router can be used. Kube-router uses the bridge plugin, but uses
+IPVS kernel module, instead of iptables. This results in better performance
+and scalability. Kube-router also uses iBGP, so that static routes are not
+requlred for pods to commincate across nodes.
```
jellonek: Two typos - s/requlred/required/, s/commincate/communicate/

pmichali: Thanks! Will fix.

pmichali: Fixing.

pmichali:
@jellonek I agree that the breakage is unrelated to the PR, and was advised by @ivan4th to comment out for now, and they can be fixed later. Has something been fixed recently, such that these will now pass? If not, then I'd just as well leave them commented out so we can get this breakage fix in. Thoughts?

pmichali:
I'm trying locally, and E2E tests seem to fail with my local repo using K8s release-1.11. Seems to still be an issue with building binaries for test run. Seems to still have an issue. I'm seeing that the test_src_release has failed in the same manner. Should I comment these two tests back out?

pmichali:
I'm trying locally, and E2E tests seem to fail with my local repo using K8s release-1.11. Seems to still be an issue with building binaries for test run. Seems to still have an issue. I'm seeing that the test_src_release has failed in the same manner. Should I comment these two tests back out?

jellonek:
The issue with building from sources should be fixed by #197 so hopefully both of these PR will be merged today.

pmichali:
@jellonek so do I wait for PR #197 and then rebase and push again?

jellonek:
Yes. That would be most proper to do. (IMO this PR could be merged without these tests, but it's up to Ivan, I would not do such thing without his approval).

pmichali:
Well, @ivan4th suggested to me, on slack, to comment out... Are we sure that PR #197 fixes source based builds? I ask, because locally, I'm seeing all sorts of issues trying to run E2E using local sources.

jellonek:
Right now on #197 they are passing but looks like they are a bit flickering... IDK if we can do something more to stabilize them on public CI...

I'm almost sure that after rebasing this PR on top of changes from #197 all the test there should be passing.

pmichali:
Can we try to quickly get PR 197 merged, so that I can quickly get this and PR 186 merged?

jellonek:
#197 is already merged so this one also should be unblocked.

PR #193, opened 2018-08-06
Fix unbound variable

jellonek:
The regression was introduced in #166. Bug is visible only when code is running without a controlling terminal.

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -1538,7 +1538,7 @@ function dind::do-run-e2e {
     fi
   fi
   dind::need-source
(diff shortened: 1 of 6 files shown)


ivan4th:
Thanks!
/approve
/lgtm

hoegaarden:
Note that the `kube_router` tests won't succeed, because `kubectl` uses the wrong / no context. This is an easy fix, I can push that later and you can rebase on that. @ivan4th @jellonek

jellonek:
@hoegaarden thx, rebased.
@ivan4th PTAL again.

PR #197, opened 2018-08-08
New syntax for kubeadm skip checks

### LINKED ISSUE #163: kubeadm-dind-cluster  fails with error --skip-preflight-checks has been deprecated
```
Flag --skip-preflight-checks has been deprecated, it is now equivalent to --ignore-preflight-errors=all
your configuration file uses an old API spec: "kubeadm.k8s.io/v1alpha1". Please use kubeadm v1.11 instead and run 'kubeadm config migrate --old-config old.yaml --new-config new.yaml', which will write the new, similar spec using a newer API version.
```

jellonek:
Use new syntax for kubeadm option to ignore checks failures for k8s
versions different than v1.8.x. Check is based on kubeadm version.

Closes #163

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -268,7 +268,6 @@ DEFAULT_DIND_LABEL='mirantis.kubeadm_dind_cluster_runtime'
 # not configurable for now, would need to setup context for kubectl _inside_ the cluster
 readonly INTERNAL_APISERVER_PORT=8080
 
-
 function dind::need-source {
   if [[ ! -f cluster/kubectl.sh ]]; then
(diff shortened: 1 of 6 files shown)


ivan4th:
Oops seems like my lgtm was wrong. Travis test case fails b/c actually using version tag for checking is incorrect. Better use `dind::kubeadm-version` like here https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/master/dind-cluster.sh#L964-L966

jellonek:
I was thinking about that initially. Ok, will use kubeadm version.

jellonek:
Code and PR description updated.

PR #199, opened 2018-08-10
Add a env variable to allow for setting of the storage mapping

rpothier:
This change will add a knob to allow for disabling of overlay2 storage mapping,
and instead use a user selected docker storage mapping.

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -300,6 +300,7 @@ SKIP_SNAPSHOT="${SKIP_SNAPSHOT:-}"
 E2E_REPORT_DIR="${E2E_REPORT_DIR:-}"
 DIND_NO_PARALLEL_E2E="${DIND_NO_PARALLEL_E2E:-}"
 DNS_SERVICE="${DNS_SERVICE:-kube-dns}"
+DIND_STORAGE_DRIVER="${DIND_STORAGE_DRIVER:-overlay2}"
 
 DIND_CA_CERT_URL="${DIND_CA_CERT_URL:-}"
 DIND_PROPAGATE_HTTP_PROXY="${DIND_PROPAGATE_HTTP_PROXY:-}"
@@ -746,6 +747,7 @@ function dind::run {
   local -a opts=("${ip_arg}" "${ip}" "$@")
   local -a args=("systemd.setenv=CNI_PLUGIN=${CNI_PLUGIN}")
   args+=("systemd.setenv=IP_MODE=${IP_MODE}")
+  args+=("systemd.setenv=DIND_STORAGE_DRIVER=${DIND_STORAGE_DRIVER}")
   if [[ ${IP_MODE} = "ipv6" ]]; then
       opts+=(--sysctl net.ipv6.conf.all.disable_ipv6=0)
       opts+=(--sysctl net.ipv6.conf.all.forwarding=1)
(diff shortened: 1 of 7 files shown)


dind-cluster.sh
```diff
@@ -815,6 +816,7 @@ function dind::run {
          --hostname "${container_name}" \
          -l "${DIND_LABEL}" \
          -v "${volume_name}:/dind" \
+         -v /tmp/${container_name}:/var/lib/docker \
```
ivan4th: This will not have the desired effect. The graph directory already resides on a docker volume mounted under `/dind` (`-v` flag above that line). See here: https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/ad68d303017cfc2d2ca198bb33995e10eaf65aa4/image/rundocker#L89
Also, creating a dir under `/tmp` on the host is not very good thing to do (a security risk among other things).

ivan4th: What's the reason for having this dir on the host as opposed to having it on the kdc's docker volume for that node?

dind-cluster.sh
```diff
@@ -1744,6 +1746,7 @@ function dind::custom-docker-opts {
     dind::step "+ Setting up insecure-registries on ${container_id}"
     jq+=("{\"insecure-registries\": ${DIND_INSECURE_REGISTRIES}}")
   fi
+  docker exec --privileged -i ${container_id} sed -i -e "s|{{DIND_STORAGE_DRIVER}}|${DIND_STORAGE_DRIVER}|" /lib/systemd/system/docker.service
```
ivan4th: It's better to pass this via `systemd.setenv` in `dind::run` maybe?

dind-cluster.sh
rpothier: This was added for running k-d-c inside a container, for k8s CI testing.
We're testing running without this.

dind-cluster.sh
rpothier: yes, changing this, thanks

image/rundocker
```diff
@@ -63,9 +65,14 @@ echo "Trying to load overlay module (this may fail)" >&2
...
-    # --storage-opt overlay2.override_kernel_check=true is needed for CentOS systems
-    storage_opts="--storage-driver=overlay2 --storage-opt overlay2.override_kernel_check=true"
+if [ "$DIND_STORAGE_DRIVER" == "overlay2" ]; then
+    DIND_STORAGE_OPTS="--storage-driver=${DIND_STORAGE_DRIVER} --storage-opt overlay2.override_kernel_check=true"
+    if grep -q '[[:blank:]]overlay[[:blank:]]*' /proc/filesystems; then
+        # --storage-opt overlay2.override_kernel_check=true is needed for CentOS systems
+        storage_opts="${DIND_STORAGE_OPTS}"
+    fi
+else
+    DIND_STORAGE_OPTS="--storage-driver=${DIND_STORAGE_DRIVER}"
```
leblancd: I think we need a setting of storage_opts here, i.e. this line:
    storage_opts="${DIND_STORAGE_OPTS}"
Note: If this line is added, this should work in a Prow DinD container if I use "DIND_STORAGE_DRIVER=vfs". If I use aufs, docker in the kube-master rejects that storage-driver type. Not sure if this is expected.

rpothier:
@ivan4th can you take another look at this? I have addressed the issues you commented on.

ivan4th:
Thanks for the PR!
/approve
/lgtm

PR #205, opened 2018-08-21
After the string split is complete, IFS should be restored

onlymellb:
After the string split is completed, the IFS should be restored to avoid affecting subsequent operations.

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -163,6 +163,7 @@ function dind::net-name {
 function dind::extract-ipv4-subnet() {
   # If only one subnet, there may be a leading space in list of subnets
   local trimmed="$( echo "$1" | sed -e 's/^[[:space:]]*//')"
+  local old_ifs=$IFS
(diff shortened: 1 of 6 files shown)


jellonek:
It looks fine, but please TAL on CLA.

onlymellb:
@jellonek  I have signed CLA and received an email from HelloSign, I don't know why this is still not passed.
image

jellonek:
@timothysc can you help somehow in this issue?

pigmej:
@onlymellb are you sure that you used the same mail to sign CLA as the one used in commits (it's not about having mail added to github account)?

pigmej:
@onlymellb then the only suggestion that I may have is:
```
If you have done the above and are still having issues with the CLA being reported as unsigned, please email the CNCF helpdesk: helpdesk@rt.linuxfoundation.org```

It seems that something went wrong during the process. I've never faced that problem before.

Issue #209, opened 2018-08-23
Add dual-stack support for k-d-c


pmichali:
Update k-d-c to support the selection of IP_MODE=dual-stack. Perform needed actions to support bringing up cluster in dual-stack mode (IPv4+IPv6).

TODO: Identify the changes needed.

pmichali:
Once PR 228 is merged, we have the basic infrastructure for dual-stack in place for k-d-c. Beyond that, there are the following enhancements that can be done:

- Test/verify and enable multi-cluster mode.
- Allow dual-stack with DNS64/NAT64 option for V6, where destination does not have IPv6 support, but using IPv6 operations.
- Additional changes made be needed, once K8s is updated to fully support dual-stack (e.g. unable to test dual-stack with Iv6 service network currently).

Issue #203, opened 2018-08-17
RFC: Alternative strategy for network defaulting


hoegaarden:
### Background

In #161 the option to run multiple clusters in parallel was introduced. Users needed to explicitly configure the subnet & external port for the kube-apiserver for every but the first cluster. To have a better UX #166 introduced mechanisms to find a free subnet for a cluster and a free port to expose the kube-apiserver on. There are some problems with the current mechanism how free networks are being found, which is why we'd like to kick of a discussion here on how other people think this problem should be solved. Sidenote: We didn't get any reports on issues with the defaulting for the kube-apiserver port, thus this mechanism will stay as is and is not part of this discussion.

### Current Mechanism

- If there is a docker network for a cluster, `DIND_SUBNET` and `DIND_SUBNET_SIZE` is queried from that docker network -> **done**
- If there is no docker network for a cluster
  - if the user has not specifically set `DIND_SUBNET_SIZE` it will be defaulted to `16`
  - if the user has not specifically set `DIND_SUBNET` it well be defaulted:
    - all docker networks are checked and the lowest IP is determined after which a network of size `DIND_SUBNET_SIZE` can be placed -> **done**
   Note: different subnet size of already existing docker networks and of the `DIND_SUBNET_SIZE` are considered and handled correctly.
  - If a user has set `DIND_SUBNET` / `DIND_SUBNET_SIZE` those settings are used -> **done**

Notes:
- With different network size (eg. `/16` vs. `/24` vs. `/18`) in play the and the current implementation the assumption that a network spans from `x.x.x.0` to `x.x.x.255` or `x.x.0.0` to `x.x.255.255` is not true anymore. Therefore getting the *n*th IP of a network cannot be done by string concatenation anymore but the IP has to actually be calculated.
- To minimise dependencies, currently no external tools are in play (e.g. `ipcalc`) but all calculation is done in shell.

### Current Problems

- This is only available for IPv4 currently. This is (the main) reason why currently multiple clusters for IPv6 are specifically disabled.
- It is relatively complicated to do IP arithmetics in shell, even more so when IPv6 comes in.
- Only the docker networks are considered. If users have any other networks on their machines, collisions might still occur.

### Suggestions

1. Fix current implementation

    Description:
    - Fix potential collisions by also excluding network (and other stuff like routes?) on the host
    - Add same mechanics for IPv6:
      - discover where a network of a certain size can be placed
[...]

PR #214, opened 2018-09-07
IPv6: Multi-cluster support and dual-stack infrastructure

pmichali:
The primary goal of this commit is to put in place infrastructure to allow
kubeadm-dind-cluster to create clusters with dual-stack networks. Because
of the needed changes, it makes sense to make sure that dual-stack and IPv6
supports multi-cluster.

A secondary goal is to attempt to maintain backward compatibility with the
current mechanisms to customize and operate the cluster.

To prepare for dual-stack, the following changes were made (there will be
follow up commits to complete support for dual-stack, some in conjuction
with Kubernetes changes that will be made):

- Introduced 'dual-stack' IP MODE and adjusted existing IPv4 only and IPv6
  only logic to continue to operate correctly, when testing $IP_MODE.

- Introduced new MGMT_CIDRS env variable to allow a comma separated list of
  CIDRs to be provided for the mangement (dind) network. For IPv4/IPv6, a
  single CIDR will be specifed.

- For backwards compatibility, DIND_SUBNET and DIND_SUBNET_SIZE, if specified
  will be merged into MGMT_CIDRS.

- For multi-cluster, a new CLUSTER_ID will be used to allow the user to
  adapt the mgmt subnet used for each cluster. Removed logic that tries to
  determine the available subnet (see note 1).

- For compatibility, if (only) DIND_LABEL is provided, a random cluster ID
  will be assigned (see note 2).

- For additional clusters, in a multi-cluster setup, resource names will
  use the suffix "-cluster-#", instead of "-<label>-<sha#>".

- The localhost address, used for API, will be based on the family of the
  service network, which will be single family, when running in dual-stack
  mode. It is unchanged for IPv4/IPv6 only, which only have one family.

- During clean, it was found that the kubectl config for the context and
  cluster (in multi-cluster) are not removed. This commit removes those
  settings.

- For IPv6 only multi-cluster, separate DNS64 and NAT64 containers are
  created, especially because the docker bridge for the mgmt network
  needs to have a IPv4 network for the NAT64 IPv4/IPv6 mapping pool.
  This IPv4 network uses the cluster ID to ensure uniqueness. These
  containers are cleaned, based on the cluster context, which is encoded
  into thier label.

- Added mulit-cluster CI test for IPv6 only (for 1.11 release).

- Modified existing multi-cluster CI test to check normal, label only
  (changed), and cluster ID (new) based clusters.

- Docker run commands are modified to define any IPs needed, based on
  the IP mode (two for dual-stack).

- Added cluster info printing, at the end of "up", so that user knows
  the network and cluster settings.

- Updated README.md to indicate the changes made.

- (FUTURE) If desired, could ensure that there is enough space in the
  mgmt CIDRs for the cluster ID to be injected.

- (FUTURE) For IPv6, could support the cluster ID only consuming 8 bits,
  instead of 16, and thus place it in either the upper or lower byte of
  the hextet, based on the CIDR size.

- (FUTURE) Could consider whether want to use /16 for mgmt CIDR in IPv4,
  allowing 32K nodes vs 254.

- (TODO) For multicluster IPv6 only, iptables rules are created on the
  host to isolate the clusters. These should be ip6tables rules. As a
  result, nodes from one cluster can currently ping nodes in another
  cluster (unless rules are manually added).

- (TODO) Complete the suport for dual-stack, including pod network, CNI,
  and other changes.

Note 1:
This came out of the need to support multiple clusters for IPv6, where
there are vast addresses and calculating an available subnet is difficult.
Picking a number to use to make the subnet unique, coupled with the base
selection via MGMT_CIDRS, provides a simple method to ensure uniqueness.

For IPv4, the management network will be /24, instead of /16, and the 3rd
octet of the subnet will be reserved for the cluster. For example, if the
MGMT_CIDRS is 10.192.0.0/24, and the CLUSTER_ID is "100", the management
CIDR will be modified to 10.192.100.0/24 for the cluster.

For IPv6, the CLUSTER_ID will be placed in the lowest hextet of the address
provided in MGMT_CIDRS (before the ::). For example, if the MGMT_CIDRS has
fd00:10:20::/64, and the CLUSTER_ID is "30", the management CIDR will be
modified to fd00:10:20:30::/64 for the cluster.

Values will be between 1..254 for IPv4 compatibility, with zero reserved for
single cluster or the first cluster in multi-cluster.

Each cluster would use a different cluster ID. The cluster ID will be used
in the names of resources, as well, to provide uniqueness.

Note 2:
Although kept for compatibility, this gives the opportunity to have a custom
cluster label, if desired. When only CLUSTER_ID is used, the cluster name will
be "cluster-#", where # is the non-zero cluster ID.

A cluster ID can be specified as well as the label, to control the mgmt
network address used.

Partially addresses issue: #209
Provides a solution for issue: #203

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -51,203 +51,308 @@ fi
 
 #%CONFIG%
 
-function dind::find-free-ipv4-subnet() {
-  local maxIP anAddressInNewSubnet
-
-  subnetSize="$1"
-
-  maxIP=$( dind::ipv4::find-maximum-claimed-ip )
-
-  if [ $maxIP -eq 0 ]
-  then
-    echo '10.192.0.0'
-    return
-  fi
-
-  # maxIP is the highest IP we cannot use (because it already belongs to a subnet)
-  # One could argue that maxIP+1 could be the MinHost of the subnet we're about to allocate (a.k.a. "new subnet").
-  # But, consider:
-  # maxIP: 10.0.0.255
-  # maybeNextMinHost: 10.0.1.0
-  # In the above, a subnet size of /16, will start at 10.0.0.0 - which is invalid.
-  # So we need to start the new subnet at least 32-(subnetSize) IP spaces away.
-  anAddressInNewSubnet=$(( maxIP + (1<<(32-subnetSize)) ))
-
-  # apply mask to get min host
-  nextMinHost=$(( anAddressInNewSubnet & $(dind::ipv4::netmask "$subnetSize") ))
-
-  dind::ipv4::itoa "$nextMinHost"
-}
-
-function dind::ipv4::netmask() {
-  local netmask i
-  netmask=0
-  for i in $( seq $(( 32 - $1 )) 32 )
-  do
-    netmask=$(( netmask + 2**i ))
-  done
-  echo "$netmask"
-}
-
-function dind::ipv4::find-maximum-claimed-ip() {
-  local maxIP upperIPs b m i
-  maxIP=0
-  upperIPs="$(
-    docker network ls --format '{{ .Name }}' | while read -r nw
-    do
-      subnet="$( docker network inspect "$nw" --format "{{ range .IPAM.Config }}{{ .Subnet }}{{ end }}" )"
-      if [ -z "$subnet" ]
-      then
-        continue
-      fi
-      IFS='/' read -r b m <<<"$subnet"
-      echo $(( $(dind::ipv4::atoi "$b") + (1<<(32-m)) - 1 ))
-    done
-  )"
-
-  for i in $upperIPs
-  do
-    if [ "$(( i - maxIP ))" -gt 1 ]
-    then
-      maxIP="$i"
-    fi
-  done
-
-  echo "$maxIP"
-}
-
-function dind::ipv4::itoa() {
-  echo -n $(( ($1 / 256 / 256 / 256) % 256)).
-  echo -n $(( ($1 / 256 / 256) % 256 )).
-  echo -n $(( ($1 / 256) % 256 )).
-  echo    $((  $1 % 256 ))
-}
-
-function dind::ipv4::atoi() {
-  local ip="$1"
-  local ret=0
-  for (( i=0 ; i<4 ; ++i ))
-  do
-    (( ret += ${ip%%.*} * ( 256**(3-i) ) ))
-    ip=${ip#*.}
-  done
-  echo $ret
-}
-
+# dind::localhost provides the local host IP based on the address family used for service subnet.
 function dind::localhost() {
-  if [[ ${IP_MODE} = "ipv6" ]]; then
+  if [[ ${SERVICE_NET_MODE} = "ipv6" ]]; then
     echo '[::1]'
   else
     echo '127.0.0.1'
   fi
 }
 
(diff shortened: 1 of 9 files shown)


pmichali:
@ivan4th I'm not sure what is up with Travis CI. It says "passed" on the details page, but seems to only have run a few tests. It say pending on this summary page. Does the test need to be restarted?

pmichali:
@ivan4th Any idea what the failures are with the src-master and src-release tests. Seems to be complaining about /k8s not being a mountpoint or invalid option.

For the IPv6 multi-cluster test case that I added, I'm not sure what I missing in the test setup. Can you advise?

pmichali:
Oh, I can rebase, but figured, I'd like to fix the test cases first.

pmichali:
Note: this latest commit has debug turned on for the multicluster tests, so that I can look at the IPv6 test case. Will remove after debugging.

pmichali:
On latest commit the three failing tests all seemed to be issues related to port forwarding. The src-master and 1.11-weave tests (inermittent) had connection refused - maybe port in use?

The multi-cluster 1.11 ipv6 seems to have the same connection refused, but I wonder if it is because port-forwarding is using localhost, instead of [::1].

Trying another temp change (in addition to the "bash -x" added for debug), to see if I can modify portforward.sh to work with IPv6 (seems like it is setup for IPv4 only). Not sure, if this is the right solution, but giving it a try. Pushed up the change.

pmichali:
With the latest change to portforward.sh, the IPv6 multicluster test is passing. The only failure is src-master (unrelated to this change set, as others' seeing on their commits), and some intermittent failures (e.g. 1.10 flannel in latest run).

Pushed patch with debug removed.

@ivan4th and others, please review the changeset.

jellonek:
/approve

@ivan4th PTAL

pmichali:
All the CI tests have passed, although today I see some as canceled, so I'm rerunning those tests (I hope). In any case, I've tested this local with IPv4 and IPv6 and am comfortable with it being out for review. PTAL folks.

dind-cluster.sh
```diff
@@ -51,204 +51,309 @@ fi
...
-	echo "WARNING! The DIND_ALLOW_AAAA_USE option is for IPv6 mode - ignoring setting."
-	DIND_ALLOW_AAAA_USE=
-    fi
-    if [[ ${CNI_PLUGIN} = "calico" || ${CNI_PLUGIN} = "calico-kdd" ]]; then
-	DEFAULT_POD_NETWORK_CIDR="192.168.0.0/16"
-    fi
+
+# Use legacy DIND_SUBNET/DIND_SUBNET_SIZE, only if MGMT_CIDRS is not set.
+legacy_mgmt_cidr=""
+if [[ -v DIND_SUBNET && -v DIND_SUBNET_SIZE ]]; then
```
ivan4th: Sadly this breaks on Mac due to old bash... Maybe `[[ ${DIND_SUBNET:-} && ${DIND_SUBNET_SIZE:-} ]]` is enough?

pmichali:
Pushed up change so that it will work with older bash for MacOS, per @ivan4th suggestion.

pmichali:
@ivan4th Have updated for MacOS and all tests passing in CI.

PR #226, opened 2018-10-03
Fix issue with kubeadm 1.13

rpothier:
With kubernetes 1.13, the Kubeadm template needs to be updated.

Fixes Issue: #223

PR #228, opened 2018-10-04
Phase 2: Dual-stack support.

pmichali:
Kubernetes is being updated to support dual-stack networking. With the
initial release, pods and management networks will have both IPv4 and
IPv6 addresses. The service network, however, will use only a single IP
family.

This commit adds (more) support to k-d-c for provisioning clusters in
this mode. The following changes have been made:

- Pod network has both IPv4 and IPv6 addresses (configurable).
- In the DinD image, the dindnet script updates include:
    * creates pair of IPs on dind0 interface.
    * configures PTP and bridge plugin with multiple subnets.
    * sets kubelet DNS arg to DNS server, if service family is V6.
    * sets node IP in /etc/hosts based on service network family.
    * sets /etc/resolv.conf entry (for DNS) same as IPv4 mode.
- Create pod network static routes for IPv4 and IPv6.
- Error checking of CIDRs provided to ensure subnet is large enough
  for suffix to be added.
- Creates ip6tables rule to provide DNAT access to outside for IPv6.
- CI test added for ipv4 and dual-stack mode (internal ping, external
  ping, namespace lookups).
- README updated to talk about dual-stack mode.
- Maintains backward compatibility with older DinD images.
- Place a /node-info file in the DinD image that indicates the image
  (not K8s) supports the configuring of a cluster in dual-stack mode.
  This file can be used for additional flags, in the future.

The changes made here, will allow a cluster to be brought up in dual-
stack mode, however, there are still changes needed in Kubernetes code
(as part of the dual-stack KEP). There are the following limitations
to this capability:

- Using IPv6 for the service network doesn't work yet. Need K8s changes.
  Have confirmed with IPv4 service network, however.
- Multicluster mode has not been tested yet for dual-stack (will add later).
- Expects that there is both IP4 and IPv6 access to the outside. This means
  that running under GCE will not work (as it expects only IPv4 access to
  outside). Future updates can add ability to enable DNS64/NAT64 to k-d-c,
  like is done with IPv6 only mode.
- It is the responsibility of the user to ensure that when they restore a
  cluster, they are using the same IP mode as when the snapshot was created.
  Since the kubeadm-dind-net network persists, after a down command, if a
  cluster is restored using a different IP mode, the network will not be
  correctly configured. This is a general limitation, unrelated to dual-stack.
- CI does not have direct access to external IPv6 hosts, so dual-stack (and
  IPv6 only) cannot test external IPv6 pings.

Currently, the test_src_master and test_1.11_dual_stack_v4_svc_net tests are
failing. The former is a know issue with the baseline. The latter needs to
be resolved (hoping for review input).

Partial Fix: #209

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -197,10 +197,46 @@ test_ipv6_only: &test_ipv6_only
         command: |
           docker load -i _save/images.tar
     - run:
-        name: Run ipv6 only tests
+        name: Run IPv6 connectivity tests
         no_output_timeout: 25m
         command: |
-          "${PWD}/test/test-ipv6-only.sh"
+          "${PWD}/test/test-net-connectivity.sh"
+
+test_dual_stack: &test_dual_stack
+  machine: true
+  working_directory: ~/kubeadm-dind-cluster
+  steps:
+    - checkout
+    - attach_workspace:
+        at: _save
+    - run:
+        name: Restore images
+        command: |
+          docker load -i _save/images.tar
+    - run:
+        name: Run dual stack connectivity tests
+        no_output_timeout: 25m
+        command: |
+          export IP_MODE=dual-stack
+          "${PWD}/test/test-net-connectivity.sh"
+
+test_ipv4_only: &test_ipv4_only
+  machine: true
+  working_directory: ~/kubeadm-dind-cluster
+  steps:
+    - checkout
+    - attach_workspace:
+        at: _save
+    - run:
+        name: Restore images
+        command: |
+          docker load -i _save/images.tar
+    - run:
+        name: Run IPv4 connectivity tests
+        no_output_timeout: 25m
+        command: |
+          export IP_MODE=ipv4
+          "${PWD}/test/test-net-connectivity.sh"
 
 test_multiple_clusters: &test_multiple_clusters
   steps:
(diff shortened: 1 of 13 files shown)


pmichali:
test_src_master is known breakage on baseline. Waiting for PR #226 to be merged. I suspect test_1.9_ipv6_only_nat64 is a flake.

test_1.11_dual_stack_v4_svc_net failure is baffling. The test should be using the 1.11 DinD image built for the test with the commit changes applied. This image, should have a /node-info file that indicates the image (and associated scripts) support dual-stack mode. However, when the cluster is brought up, the image is checked and the file is missing (so the `up` is aborted and the test fails). I'm not sure why the test is not picking the right image (or the image is missing the needed file). Ideas?

pmichali:
OK, tests were a bit flakey (had to rerun 2x, as was getting an error when script was doing daemon-reload of kubelet). Should be all set for review now.

pmichali:
Looks like I must have accidentally closed, when I did my last comment. Opened back up. This is ready for review and has been tested locally and in CI.

pmichali:
@ivan4th @pigmej PTAL...

pmichali:
@ivan4th Thanks for the review!!!

PR #233, opened 2018-10-09
Allow dind-cluster[-v1.11].sh to run on the latest Docker for Mac

mestery:
The latest versions of Docker for Mac have changed what directories by default are
setup for bind volume mounts. The dind-cluster*.sh scripts assume that /boot and
/lib/modules can be bind mounted, but this is not the case with the new Docker for
Mac versions.

This commit detects if the script is running on a Mac, and if so, it replaces /boot
and /lib/modules with /tmp, which is by default bind mountable on the latest Docker
for Mac.

Tested on Docker for Mac version 2.0.0.0-beta1-mac75 (build 27117) with Kubernetes
version 1.11.0.

Signed-off-by: Kyle Mestery <mestery@mestery.com>

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -1022,7 +1022,7 @@ function dind::run {
   dind::step "Starting DIND container:" "${container_name}"
 
   if [[ ! ${using_linuxkit} ]]; then
-    if [ "${DARWIN_MODE}" == "True" ] ; then
+    if [ "${DARWINMODE}" == "True" ] ; then
       # /boot and /lib/modules do not exist with Docker for Mac, but
       # /tmp is available for bind mounting.
       BOOTVOL=/tmp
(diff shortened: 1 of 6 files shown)


mestery:
I've signed the CLA, replying with a comment per the instructions above.

mestery:
This PR has been open for 27 days now, can I get a maintainer to eyeball this please? Also, it's unclear to me why the `tide` job is still pending, so any guidance there would be appreciated.

cscetbon:
Hey @mestery, I should have taken more time to look for your PR. I made another one to do something similar https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/246. I don't think you even need to set those variables. Just saying it's not a linux system should be enough. At least it works fine on my machine. what do you think ?

mestery:
@cscetbon I like your approach better, I'll give it a try tomorrow, but it LGTM from eyeballing it.

Issue #236, opened 2018-10-15
k-d-c failing with newer KubeAdm versions


pmichali:
With newer KubeAdm versions, we are seeing that the kubeadm.conf file has significantly different format, which is difficult to accommodate with the current k-d-c code.

Should have version specific kubeadm.conf files, using templates, instead of trying to morph a single template to work for multiple versions.

I'll be pushing up a PR in a few minutes.

pmichali:
There are two things seen. First, the API version for 1.13 is incorrect in the config file created (alpha3 vs beta1). Version 1.12 uses alpha2 (still compatible) instead of alpha3.

Second, because there are considerable changes to kbueadm.conf syntax from version to version, it is difficult to customize the template used, because of considerable syntax difference. 

There is a desire (at least by me) to have a config file that is more representative of the format used for the release, so that it can be customized, if desired, and so that getting support from the Kubernetes community is easier (the first thing they ask for is the config file, and the first thing they comment about is that it is not the right format).

To get to the right format for a release, one can run the kubeadm config migrate command, but that must be done on each release (install 1.11, run migrate from 1.10 to 1.11, install 1.12, run migrate from 1.11 to 1.12, ...). Tedious would be a nice term for this.

PR #237, opened 2018-10-15
Version specific kubeadm.conf files

pmichali:
Changed this to use version specific config files and to use them
based on the kubeadm version, rather than having one template file
and trying to morph it for each version. Foudn that 1.12 is
considerably different config.

The 1.10 template is used for 1.8-1.10, and the 1.13 template is
used for 1.13 and newer versions.

Note: There are new sections and fields for the 1.12 an 1.13 kubeadm.conf
files. For 1.13, the conntack settings were commented out, as I suspect
they are host specific. Other, like the healthz ports, were left there,
even though they are showing IPv4 addresses. This template was generated
from a IPv6 setup, so, at this time, IPv4 is still used for some
things. These may need to be parameterize or removed later, as IPv6
changes are made.

In addition, a problem was seen during kube-router cleanup operations,
causing script to fail with latest kube-router code. Made change to
in-line include a version pinned kube-router config file and call out
that version during cleanup as well.

In future, this could be removed, or the version (and config) could be
updated, as newer kube-router versions are released.

Fixes Issue: #236


pmichali:
@ivan4th Please take a look. This has per-version config files for kubeadm, and nails kube-router to a specific version.

lukaszo:
@pmichali can you split it into two PRs? One for kubeadm and the second one for kube-router? The second one will unblock other PRs

pmichali:
OK. Created new issue and splitting out the PRs.

pmichali:
@ivan4th @lukaszo rebased to get kube-router fix, and pushed up. Please review.

lukaszo:
I'm not brave enough to merge it without @ivan4th review :) 

/approve

pmichali:
@ivan4th Can you review please?

ivan4th:
Works for me. Thanks!
/approve
/lgtm

pmichali:
Thanks Ivan!

PR #238, opened 2018-10-16
Install CNI_PLUGIN="weave" via cloud.weave.works

brb:
This is the official way to install Weave Net as shown at https://www.weave.works/docs/net/latest/kubernetes/kube-addon/#install.


lukaszo:
@brb can you rebase? kube-router failures should be fixed in current master

brb:
@lukaszo Rebased.

lukaszo:
It looks like `/retest` doesn't work here. The failed test is not related to the change.

/lgtm

PR #246, opened 2018-11-02
fix_start_on_mac - Do not try to use /boot on Mac

cscetbon:


mestery:
@cscetbon You also need to sign the CLA, and then comment back, to get the CI to pass cleanly.

mestery:
@cscetbon Also note the `tide` check never seemed to run on my PR, and it sat there for 27 days, so unless a maintainer actually responds here, that may stick in `pending` state until the dawn of time ...

lukaszo:
@mestery sorry for not checking your PR
@cscetbon you need to sign cla, without it we can not merge ie

mestery:
@lukaszo Any idea why the `tide` CI run never seems to complete?

ivan4th:
This PR seems to be using not quite correct approach, disabling the mounts of `/boot` and `/lib/modules` on all the platforms. Also, I can't reproduce the original issue on my Mac. Could you please paste `docker info` output and also `bash -x ./dind-cluster-v1.11.sh` output on the unchanged k-d-c? Thanks.

cscetbon:
> This PR seems to be using not quite correct approach, disabling the mounts of /boot and /lib/modules on all the platforms.

@ivan4th what do you mean ? It disables it by default and enables it on anything else than Windows/Mac

cscetbon:
@lukaszo cla signed

ivan4th:
Ok I see. Perhaps you're using edge Docker and because of this you have this line
in Docker Info:
```
Kernel Version: 4.9.125-linuxkit
```
whereas the check for linuxkit suffix in k-d-c looks like this
```
docker info|grep -s '^Kernel Version: .*-linuxkit-' > /dev/null 2>&1 ; then
```
(note the hyphen `-` after `linuxkit`). The check works for my Mac Docker
version for example:
```
docker info|grep Version
Server Version: 18.06.1-ce
Kernel Version: 4.9.93-linuxkit-aufs
```
but not yours.

The idea of the script is **NOT** to mount `/boot` and `/lib/modules` when `using_linuxkit` is set to a non-empty string (e.g. `1`), b/c `linuxkit` is what Mac Docker is using.
Note that in the `bash -x` run above, `using_linuxkit` is never set to a non-empty string.

The mistakes in your patch are (made comments for corresponding lines before that, too):
1) setting `using_linuxkit=0` will prevent the script from mounting `/boot` and `/lib/modules` in any case because a check which is used below this line `[[ ! ${using_linuxkit} ]]` is always false for any `using_linuxkit` value except an empty string, and `0` is not an empty string;
2) adding `|Mac` to Windows check clause. Actually, now after looking at your PR I suspect that #83 wasn't correct either and Windows docker must be fixed using another approach, but I don't have any Windows machines to check this. In this case, this problem doesn't change anything b/c `using_linuxkit` is always true due to 1.

The proper fix is actually just removing the second hyphen `-` from `linuxkit` grep.
Please try this patch and feel free to replace this PR with it if it works for you:

https://pastebin.com/mr351Dt4

ivan4th:
As of `tide` check, it'll "pass" after the PR gets `lgtm` label.

cscetbon:
@ivan4th okay this time we're on the same page. I pushed the fix

ivan4th:
Thanks and sorry for the delays.
/lgtm

cscetbon:
@ivan4th some pending tasks actually succeeded and the error can't have been caused by that minor change.

ivan4th:
Agree, no sense in restarting the jobs I think. Merging.

PR #263, opened 2018-11-29
Fix custom docker opts

mattmb:
Currently setting custom docker opts seems to be broken. I get:
`Job for docker.service canceled.` and then an exit.

I think this is because we are doing `service docker restart` before we
do:
`Created symlink
/etc/systemd/system/multi-user.target.wants/docker.service →
/lib/systemd/system/docker.service.`

My proposed fix is just to ignore the failed restart. Not really ideal
but I can't work out how to re-order it? Maybe someone more familiar can
suggest a better solution.


ivan4th:
Sorry for delay. You need to run `build/genfixed.sh` and include fixed scripts in the commit.

mattmb:
Thanks, I've updated my commit anyway but it looks like @jc-sanchez has a better fix in https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/271 so feel free to close this if you merge that PR.

pigmej:
Closed because of #271

PR #273, opened 2018-12-19
Add support for k8s 1.13 and introduce releases

### LINKED ISSUE #255: dind-cluster does not work anymore
The latest image seems broken. A week ago I was able to set up a k8s cluster but now anytime I try I get 

```
* Making sure DIND image is up to date
v1.11: Pulling from mirantis/kubeadm-dind-cluster
Digest: sha256:6901087e83e9b04469a4d723e6582aefa80589ddd3137dc54449e4824005488c
Status: Image is up to date for mirantis/kubeadm-dind-cluster:v1.11
* Removing container: 39db1a014f6e
39db1a014f6e
* Starting DIND container: kube-master
A dependency job for docker.service failed. See 'journalctl -xe' for details.
```

I use Docker for Mac CE Edge channel. I also tried on a Linux box and got the exact same issue. I checked with both 1.11 and 1.12. 

When I manually start the docker service in the kube-master container I get 

```
# systemctl status docker.service
WARNING: terminal is not fully functional
● docker.service - Docker Application Container Engine
   Loaded: loaded (/lib/systemd/system/docker.service; disabled; vendor preset: enabled)
   Active: inactive (dead)
     Docs: https://docs.docker.com

Nov 23 01:18:16 kube-master systemd[1]: Starting Docker Application Container Engine...
Nov 23 01:18:16 kube-master systemd[1]: Dependency failed for Docker Application Container Engine.
Nov 23 01:18:16 kube-master systemd[1]: docker.service: Job docker.service/start failed with result 'dependency'.
Nov 23 01:18:16 kube-master systemd[1]: Stopped Docker Application Container Engine.
```

### LINKED ISSUE #262: bad form - please don't change versioned code from under us
Typically when you stamp some code with a non snapshot version, like v1.12, it becomes immutable.  I won't say it's religion, but certainly a pretty widely accepted thing.

A few days ago the `dind-cluster-v1.12.sh` script changed, causing a miss match of script and docker images, which instantly failed anything that depended on kubeadm-dind-clusters. A minor inconvenience to be sure, I just needed to update our version of the script, but none the less broke a sacred rule of devops..
>_thou shall not double release the same version_

(...hey I didn't say it wasn't my religion)

Tools like artifactory don't even let you push a release twice exactly because of this. So I humbly ask, can we either never change locked in release versions of the dind-cluster scripts in master, or start doing releases in this repo?  And docker images with release version tags should never change... ever. Please just cut a new version? Or, can we start using snapshots or some other non release identifier?

I would be happy to help anyway I can

EDIT
Just to be clear, I am suggesting decoupling kubeadm-dind-cluster version from kubernetes version so we can lock in immutable things... ie

dind-cluster-v1.12.sh -> dind-cluster-v1.12-0.1.0.sh

ivan4th:
With adding support for k8s 1.13, k-d-c will switch to using GitHub releases
instead of `fixed/` script. Each released script will be pinned to its particular
image tagged with commit id and k8s version, and with added SHA256
digest, so issues with script-image mismatch will no longer happen.
It's also possible to use images produced for non-released commits
by means of setting `DIND_COMMIT=kdc_commit_id_for_your_copy_of_dind_cluster_sh`.

The release is to be made after this PR  is merged. Also, `fixed/` directory
will be updated from the release for backward compatibility (this will be
happening with each release for a while).

The solution is more or less outlined in this comment:
https://github.com/kubernetes-sigs/kubeadm-dind-cluster/issues/255#issuecomment-442415465
except that tag/branch names aren't used for image tagging, just commit & k8s version.

This PR also introduces auto-generated test matrix for CI and improves the handling of mounts in control plane pods for k8s 1.11+.

Fixes #262 
Fixes #255


build/build-all.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+# Copyright 2017 Mirantis
```
jellonek: 2018

README.md
```diff
@@ -49,39 +51,44 @@ IPv6 and thus clusters cannot be formed using IPv6 addresses.
...
 
 $ # add kubectl directory to PATH
 $ export PATH="$HOME/.kubeadm-dind-cluster:$PATH"
 
 $ kubectl get nodes
 NAME          STATUS    ROLES     AGE       VERSION
-kube-master   Ready     master    4m        v1.12.1
-kube-node-1   Ready     <none>    2m        v1.12.1
-kube-node-2   Ready     <none>    2m        v1.12.1
+kube-master   Ready     master    4m        v1.13.1
```
jellonek: v1.13.0? 
In build/buildconf.sh there are hashes only for it, not for 1.13.1.

image/kubeadm.conf.1.11.tmpl
```diff
@@ -7,9 +7,9 @@ api:
   bindPort: 6443
```
jellonek: Missing space/non standard indent level.

image/kubelet.service
```diff
@@ -11,7 +11,7 @@ Environment="KUBELET_EVICTION_ARGS=--eviction-hard='memory.available<100Mi,nodef
 Environment="KUBELET_DIND_ARGS="
 Environment="KUBELET_LOG_ARGS=--v=4"
 # FIXME: this breaks k8s 1.10
```
jellonek: Is this FIXME still needed? If so - is it breaking 1.10 build?

build/build-all.sh
ivan4th: Done.

README.md
ivan4th: Fixed, thanks.

image/kubeadm.conf.1.11.tmpl
ivan4th: The indent differences (4 spaces below this place) are due to template differences between 1.13 and other versions. Added a comment for that. On this line, there are 2 spaces.

image/kubelet.service
ivan4th: The FIXME is no longer needed as this line is removed in case of 1.10.
Thanks, removed.

kkimdev:
Thanks for the fix! Though one thing: the link mentioned in README.md https://github.com/kubernetes-sigs/kubeadm-dind-cluster/releases/download/v0.1.0/dind-cluster-v1.13.sh gives 404

ivan4th:
@kkimdev right now there'll be short transitional period (2-3h) after which the release will happen
and `fixed/` directory will be re-populated. The link will become live once the release workflow
completes on CircleCI, if all goes well https://circleci.com/workflow-run/ae8b678d-5e4b-474f-94a6-3fbe89016126

kkimdev:
@ivan4th Oh I see.  Thanks!!

PR #293, opened 2019-03-13
Fix compatibility with current master (1.15-alpha.0)

ivan4th:
MountPropagation is no longer a feature gate.
Also, had to adjust grabbing the "join" command
b/c of the new control plane join feature.


jellonek:
Btw. shouldn't that be distinguished between different versions and kept as is for older versions?

ivan4th:
@jellonek thanks for lgtm! As for the fix, 1.10 is on its way out, the support for it will be removed once k8s 1.14 is released.

Issue #289, opened 2019-03-07
Old kubernetes dashboard deployment


morningspace:
Hi,

I noticed the current version of the default [deployment](https://rawgit.com/kubernetes/dashboard/bfab10151f012d1acc5dfb1979f3172e2400aa3c/src/deploy/kubernetes-dashboard.yaml) for kubernetes-dashboard is v1.6.0 which is a bit old. Guess it's probably because the dashboard enforced its security and can only be accessed over HTTPS by default since v1.7?

To live with an old dashboard may be inconvenient in some cases. Although the dashboard URL can be customized, there's a *bug* if I choose newer version. When call `dind::component-ready` to detect the dashboard readiness, it uses an old label `app` in 1.6 which has been changed to `k8s-app` after then.

To have a default dashboard URL for people who don't care it is great. However, can it support new version as well in case some people like me may need it? (`app` vs. `k8s-app`)  :-)

Also, according to rawgit.com, the site will be shut down soon. There needs to be a new place to put the deployment YAML for the old one if that's needed by people. Alternatively, we could even disable dashboard deployment by default, and let people customize `DASHBOARD_URL` by themselves explicitly. Seeing someone was asking [a similar feature](65) while ago.

I'd be glad to contribute code if that makes sense!

morningspace:
Hi, I've submitted a PR for this at #294 . It also addresses #65 Feel free to comment on it...

morningspace:
I am gonna close it since the PR #294 has been merged! Cool...

Issue #65, opened 2018-01-08
Don't force dashboard installation


nhoughto:
For our use case we don't require the kubernetes dashboard, so it would be useful for us (and save time) to have a flag that would not require it to be installed and wait for it to come up before completing a run / restore.

PR #294, opened 2019-03-13
Add SKIP_DASHBOARD flag that allows to deploy dashboard optionally

morningspace:
This PR is for #289 and #65

ChangeLog:
* Add SKIP_DASHBOARD and do not deploy dashboard when it's set to 1. By default, it'll be deployed.
* Break up the code that wait for DNS service and dashboard be ready, and extract the wait-for logic into a util method.
* Support both new and old version of the dashboard where the label is different: `app` for 1.6 vs. `k8s-app` for later version.
* Print Access dashboard info both for HTTP and HTTPs
* Break up the code in `snapshot` to scale down DNS service and dashboard when in prepare method. Only handle dashboard when it's detected to be there.


morningspace:
I signed it

ivan4th:
Thanks for the PR!
/lgtm

Issue #295, opened 2019-03-14
Allow cluster nodes to join custom network(s) besides the default kubeadm-dind-net


morningspace:
When we launch cluster, all nodes will join the default `kubeadm-dind-net`, which is the management network for the cluster. The basic idea here is to allow nodes to join custom network(s) other than the default one. This allows nodes to connect to something outside of `kubeadm-dind-net`.

One typical use case that I'm doing is to add these nodes to a local bridge network where there are a few private Docker registries which are also containers, so, when launch the cluster, kubeadm can pull images from these private registries quickly rather than their public peers, e.g. `k8s.gcr.io`, `gcr.io`, etc. I found that makes the cluster bootstrap really really fast, both for local dev and CI. And, this can even run **w/o** network connection since all traffic happens at local...

I'm going to submit a simple PR for this... Feel free to check it and see if that makes sense to merge when people think that's useful to others...

morningspace:
closing it as the PR has been merged

PR #296, opened 2019-03-14
Allow nodes to join custom network(s) during cluster launch

morningspace:
This PR is for #295 

ChangeLog:
* Add `DIND_CUSTOM_NETWORKS` flag that allows people to specify comma-separated custom networks, typically the user-defined Docker bridge networks.
* Run `docker network connect` right after each containerized cluster node is launched if `DIND_CUSTOM_NETWORKS` is specified.

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -524,8 +524,6 @@ DIND_NO_PROXY="${DIND_NO_PROXY:-}"
 DIND_DAEMON_JSON_FILE="${DIND_DAEMON_JSON_FILE:-/etc/docker/daemon.json}"  # can be set to /dev/null
 DIND_REGISTRY_MIRROR="${DIND_REGISTRY_MIRROR:-}"  # plain string format
 DIND_INSECURE_REGISTRIES="${DIND_INSECURE_REGISTRIES:-}"  # json list format
-# comma-separated custom network(s) for cluster nodes to join
-DIND_CUSTOM_NETWORKS="${DIND_CUSTOM_NETWORKS:-}"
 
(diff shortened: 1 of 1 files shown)


morningspace:
Anybody know what's happening. I'm seeing the `test_src_master` build keeps failing even I revert my changes... From the log, it looks the latest kubernetes (v1.15.0) built from source cannot be brought up successfully. It always fails at:
```
[wait-control-plane] Waiting for the kubelet to boot up the control plane as static Pods from directory "/etc/kubernetes/manifests". This can take up to 4m0s
[kubelet-check] Initial timeout of 40s passed.
[kubelet-check] It seems like the kubelet isn't running or healthy.
[kubelet-check] The HTTP call equal to 'curl -sSL http://localhost:10248/healthz' failed with error: Get http://localhost:10248/healthz: dial tcp 127.0.0.1:102
```

While I can see it pass on my local, with the same steps... is that something wrong w/ the CI env where I couldn't touch...

@ivan4th @roberthbailey

morningspace:
@ivan4th As all tests passed now, would you mind taking look and see if that makes sense to merge... thanks...

ivan4th:
Thanks for the PR.

/approve
/lgtm

ivan4th:
Hmm, LGTMd it before noticing that the commits could be squashed, as the change itself is ok, tried to delete my comment but too late :) 

Ok, it'll have to stay this way...

morningspace:
Ah...I should have done that...Anyway thanks for reminding that and I will definitely do that next time :-)

PR #297, opened 2019-03-14
fix: dind-cluster script pattern match comparison failures

### LINKED ISSUE #292: dind-cluster.sh scripts use pattern match comparison
this line of e.g. the 1.12 script: https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/master/fixed/dind-cluster-v1.12.sh#L1102

copied here:
>  if [[ ${DIND_LABEL} = ${DEFAULT_DIND_LABEL} ]]; then

uses a "pattern match" style bash comparison, as indicated by the lack of any quotation marks. see here for discussion: https://unix.stackexchange.com/a/72042

was this intentional? i note that the dind script uses bitwise comparisons of these variables in two other places, e.g.
https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/master/fixed/dind-cluster-v1.12.sh#L75
https://github.com/kubernetes-sigs/kubeadm-dind-cluster/blob/master/fixed/dind-cluster-v1.12.sh#L548

i'm asking because starting last night, we began to get failures in travis on line 1102, syntax errors. changing that line to read
>  if [[ ${DIND_LABEL} = "${DEFAULT_DIND_LABEL}" ]]; then

seems to have resolved the problem. by adding quotes, we tell bash to use a string/bitwise comparison rather than a pattern match comparison.

starpit:
this PR changes a pattern-match based comparison of variables to a string-based comparison. this resolves issues with use of kubeadm-dind within e.g. travis

Fixes #292

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -1097,7 +1097,7 @@ function dind::configure-kubectl {
     --server="http://${host}:$(dind::apiserver-port)" \
     --insecure-skip-tls-verify=true
   "${kubectl}" config set-context "$context_name" --cluster="$cluster_name"
-  if [[ ${DIND_LABEL} = ${DEFAULT_DIND_LABEL} ]]; then
+  if [[ ${DIND_LABEL} = "${DEFAULT_DIND_LABEL}" ]]; then
       # Single cluster mode
       "${kubectl}" config use-context "$context_name"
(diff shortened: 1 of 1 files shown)


starpit:
CLA work done

Issue #299, opened 2019-03-20
test_src_master build fails due to recent change from Kubernetes src


morningspace:
I am seeing test_src_master build keeps failing recently and it's stopped at node join as below:
```
Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /lib/systemd/system/kubelet.service.
Error: accepts at most 1 arg(s), received 2
Usage:
  kubeadm join [api-server-endpoint] [flags]
  kubeadm join [command]
```

Then noticed that [the recent fix](/kubernetes-sigs/kubeadm-dind-cluster/commit/30225fa2d45075055a5854cbe00ac1fcded11a06#diff-6f53bfa951dee02d96a2466da8374ee8R1323) to grab the join command flags never works after [this change](/kubernetes/kubernetes/commit/fcadf143a5ab99ea4318caad22815091e793f486#diff-7c6f7ac7c8c8a0ecc24b1ade8fde8c34L33) went into Kubernetes code repo just a couple of days ago.

I'm going to submit a PR to fix it.

morningspace:
am closing it as it's merged... cheers...

PR #300, opened 2019-03-20
Fix the test_src_master build failure

morningspace:
This is to fix the issue #299 

ChangeLog:
* Update the code to grab the join flags from kubeadm output according to the latest change introduced into Kubernetes source.

The original line was split into two by the author,  so that it needs to grab both the line matched and the line right after that. Also, it needs to drop the trailing backslash and carriage return.

P.S.
Although the approach to grab is improved here, and I did test some other possible variant outputs, it appears to me that to use grep/sed to grab the flags from the output is generally unreliable. Also, it depends on how often the output of these flags are changed :-)

### DIFF
--- a/dind-cluster.sh
+++ b/dind-cluster.sh
@@ -1322,7 +1322,7 @@ EOF
   if [[ ${BUILD_KUBEADM} || ${BUILD_HYPERKUBE} ]]; then
     docker exec "$master_name" mount --make-shared /k8s
   fi
-  kubeadm_join_flags="$(dind::kubeadm "${container_id}" init "${init_args[@]}" --ignore-preflight-errors=all "$@" | grep 'kubeadm join.*--token' | tail -1 | sed 's/^.*kubeadm join //')"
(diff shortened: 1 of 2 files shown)


morningspace:
Also added one more code change in `snapshot`, as I'm seeing there's a controller manager glitch issue happened intermittently during the CI runs. I'd hope that's reviewed and merged together w/ the fix to test_src_master build failure issue, or I can open another PR if needed.

The reason for the change to `snapshot`: my recent change made to `snapshot` in #294 is to scale down the DNS service first, then the dashboard if it's there. This is probably problematic when check the dashboard given the DNS is down. While the original code is to scale down the two services first, then detect their status. What I modified in this PR is using exactly the same approach. This will make the process much more reliable.

morningspace:
cc @ivan4th

morningspace:
I see all CI builds passed... that's cool!

ivan4th:
Thank you!
/approve
/lgtm

PR #308, opened 2019-05-25
Update example to release 0.2.0 and Kubernetes 1.14

npalm:
I have updated the example in the README. Made the example slightly more generic so updates for future release would be easier. Bumped the version of kubeadm-dind-cluster to the latest release tag 0.2.0 and set Kubernetes to 1.14.


npalm:
@pigmej I have only updated the README. But for some reason that is not clear to me the build fails.

PR #312, opened 2019-07-11
Adding support for version 1.15

wk8:
Also updated other minor versions to their latest respective patches.

Added a way to make the kubelet's flags depend on its version, too: that
is needed since v1.15 does not have the `allow-privileged` flag any
more, but we still want to have it present for older versions; made that
part of `dind::init`.


wk8:
Tests fail because the kubelet 1.15 won't accept the `--allow-privileged` flag. Will fix this PR in the very next days.

PR #313, opened 2019-07-12
Fixing linuxkit detection on the latest versions of Docker for Mac

wk8:
Recent versions of Docker can have leading spaces in the output of
`docker info`, which the regexes used to detect linuxkit didn't account
for. Fixing that, and also consolidated the calls to `docker info` as it
can take half a second every time to return.

Signed-off-by: Jean Rouge <rougej+github@gmail.com>


wk8:
The test failure is unrelated, as evidenced by the same failure appearing on https://github.com/kubernetes-sigs/kubeadm-dind-cluster/pull/314, which doesn't change anything.

PR #315, opened 2019-07-15
Add pause/unpause options

cscetbon:
Pausing docker containers is useful when the cluster sucks up resources and we don't want to take it down. For example there is no way to take more than one snapshots and taking down/up the cluster can take time.


cscetbon:
@ivan4th please check CircleCI, it seems to come from timeout issues not related to my new functions

